PR #1, opened 2013-10-08
add the supports of vcs tag

janncker:
Signed-off-by: Hasan Wan hasan.wan@intel.com

### DIFF
--- a/src/package.c
+++ b/src/package.c
@@ -170,6 +170,7 @@ cr_package_copy(cr_Package *orig)
     pkg->arch             = cr_safe_string_chunk_insert(pkg->chunk, orig->arch);
     pkg->version          = cr_safe_string_chunk_insert(pkg->chunk, orig->version);
     pkg->epoch            = cr_safe_string_chunk_insert(pkg->chunk, orig->epoch);
+    pkg->vcs              = cr_safe_string_chunk_insert(pkg->chunk, orig->vcs);
     pkg->release          = cr_safe_string_chunk_insert(pkg->chunk, orig->release);
     pkg->summary          = cr_safe_string_chunk_insert(pkg->chunk, orig->summary);
     pkg->description      = cr_safe_string_chunk_insert(pkg->chunk, orig->description);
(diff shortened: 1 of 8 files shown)


Conan-Kudo:
@Tojaj If I might make a suggestion? These extended attribute tags should probably be behind a switch/config option so that they aren't on by default. I definitely agree that the repodata should be as minimal as possible, but @kad's point that Debian does include it in their metadata (for better or worse) means that other people may in the future find it useful. 

I'm not entirely sure why the Tizen Project needs this information in the metadata for the repository, but provided that it's not a "default on" feature, I don't see any reason why it can't be there.

kad:
In Tizen it was used for very specific reason - to be able to track from binary RPM distribution back to origins of source, as VCS tag includes full URL and commit id of sources used to generate package (both actual sources and corresponding meta-data, with help of support git-buildpackage).

Conan-Kudo:
@Tojaj Could this be brought in if the generation of this was triggered by a flag rather than default behavior? I'd rather not see the metadata get bigger (I would rather see it slimmed down and support xml.xz format to be even smaller), but Tizen is using this tag, so I don't see a good reason for it not to be in there, again, as long as it isn't pulled in by default.

kad:
Scenario where it would be included only if additional switch specified to createrepo as described by @Conan-Kudo would be good enough solution in my opinion.

Conan-Kudo:
@janncker Do you think you could implement what I suggested earlier for your pull request? It would probably make it much more palatable to merge in.

janncker:
@Conan-Kudo I can help on this even I'm not on this project now.  But, after a quick search on the latest branch, there's no mechanism to make a tag optional. @Tojaj do you have plan to support tags as optioinal? Or global cmd option is acceptable?

Conan-Kudo:
@janncker I suspect that a global command option that would enable the functionality (default disabled) would be fine.

j-mracek:
I am sorry but I am closing the PR as an inactive.

PR #11, opened 2015-01-24
Bring bash completion install dir and filenames up to date with current bash-completion

scop:
Note that the correspond rpm specfile (e.g. Fedora one) will need a build dependency on bash-completion as well as adjustments for the dir and file names.


Tojaj:
Thanks!

PR #13, opened 2015-02-06
Python: Get the updateinfo generation examples working.

lmacken:
Prior to this patch, both Python updateinfo generation examples yield:
```
    Traceback (most recent call last):
      File "updateinfo_gen_02.py", line 56, in <module>
        generate()
      File "updateinfo_gen_02.py", line 38, in generate
        rec.issued_date = "2014-08-14 00:00:00"
    TypeError: DateTime or None expected!
```


Tojaj:
Thanks!

PR #14, opened 2015-02-08
Examples: Add a simple modifyrepo python example

lmacken:
I just ported bodhi over to createrepo_c (https://github.com/fedora-infra/bodhi/pull/128), and in the process wrote a basic Python example of inserting `updateinfo.xml` into repodata like `modifyrepo`

### DIFF
--- a/examples/python/simple_modifyrepo.py
+++ b/examples/python/simple_modifyrepo.py
@@ -1,0 +1,35 @@
+#!/usr/bin/env python
+"""
+An example of inserting updateinfo.xml into repodata.
+"""
+import os
+import shutil
+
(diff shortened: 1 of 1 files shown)


Tojaj:
Thanks, I merged the pull request, but I did a small refactoring: https://github.com/Tojaj/createrepo_c/commit/3d9d796958de2ac057b8cb884383636f96421dd3

Tojaj:
I did the refactoring, because you example takes a GZziped updateinfo file, compress it again with XZ and then adds it to the repodata. After my refactoring, the example takes the GZziped file and add it to repodata without a redundant compression.

PR #34, opened 2015-08-20
updateinfo: Fix a typo in the package release attribute

lmacken:
I think this issue is causing yum to complain:

https://github.com/fedora-infra/bodhi/issues/275


Tojaj:
Thanks!

PR #35, opened 2015-09-17
fix traceback on non-complete datetime information (python)

jaroslawp:
Some Red Hat advisories contain incomplete datetime information in "issued date" and/or "updated date" updateinfo.xml elements - YYYY-MM-DD instead of expected YYYY-MM-DD HH:MM:SS
attached patch prevents python Traceback on parsing such data.

Please consider adding the patch.

Best Regards

Jarek


Tojaj:
Thanks!

PR #37, opened 2015-09-20
Add missing compile-time definition to enable DRPM support

Conan-Kudo:
Previously, even with being linked with libdrpm, it would still fail to properly enable DeltaRPM generation capabilities in createrepo_c. This was because no compile-time definition was being passed to enable those blocks of code throughout the source files.

This PR fixes that. However, one other fix is required that can't be covered by this PR. In the RPM spec, you'll need to fix your BuildRequire for `drpm` to point to `drpm-devel`, otherwise there will be no header information for libdrpm for the code to use.


Tojaj:
Hi Conan,
thanks for all the investigation! Good work!
In the end, I patched this bug by most suitable solution, directly in main CMakeLists.txt:
3c97a65e210a8505adb9d95df1b1a78fa38812b1
because just few lines above, this is done for manually specified drpm paths:
https://github.com/rpm-software-management/createrepo_c/commit/3c97a65e210a8505adb9d95df1b1a78fa38812b1#diff-af3b638bc2a3e6c650974192a53c7291L76

I'll fix the spec file in rawhide fedora soon.

Thanks
Tomas

Conan-Kudo:
@Tojaj the set() command does not set up a definition to be passed to the compiler. You need to use add_definitions() to make it pass it to the compiler. You already had the set() command, and it does nothing.

Tojaj:
I know that it only sets a var and that's the purpose :)
I need to set the cmake variable `CR_DELTA_RPM_SUPPORT` as the var is being placed into `deltarpms.h`.
The `deltarpms.h` is generated from `deltarpms.h.in` during call of `CONFIGURE_FILE` from `src/CMakeLists.txt`.
The set does something, if you would try to build createrepo_c with explicitly set drpm path, you will see that the set work, but implicitly detected drpm didn't set this var and that was the problem.

But thanks for double checking :+1:

Conan-Kudo:
@Tojaj If I understand this correctly, it uses `/usr/bin/makedeltarpm`, so you should probably define the deltarpm package as a "Requires" for createrepo_c-libs or wherever it makes sense to ensure that the binary can be found. Unless you're now using the new libdrpm DeltaRPM creation code?

PR #39, opened 2015-11-08
Fix manpage generation and install to create and transfer uncompressed man pages

Conan-Kudo:
You currently provide pre-generated, pre-compressed man pages for your CMake build script to install.

There are a couple of problems with this:
* Because they are pre-compressed, they get decompressed and recompressed at package build
* Because they are binary, they are not properly versioned in git

This pull request adjusts it to generate and install uncompressed man pages. The compression will be taken care of by tools like rpmbuild anyway. Additionally, the man pages are updated to match the latest release.


Tojaj:
Thank you!

PR #43, opened 2015-11-16
Python3 support.

### LINKED ISSUE #41: [RFE] Provide a Python 3 API
Currently createrepo_c offers a Python 2 API, but it does not offer a Python 3 API to go with it. 

Please add a Python 3 API to createrepo_c for those who want to have their programs use Python 3 instead of Python 2.

ralphbean:
I was able to get the test suite passing with both Python2 and Python3.  :)

Notable things that changed:

- Strings are now handled explicitly as either Bytes or Unicode, which is a requirement for Python3, but is nice to have anyway on Python2.
- The interface for initializing modules changed, so some conditionals were used to span the two versions.
- The old interface for declaring types had been deprecated, and is removed in Python3.  The new interface works with Python 3.x as well as Pythons 2.6 and 2.7.  If we need Python 2.5 support, we'll have to go back in and add some more preprocessor conditionals there.

Let me know if anything is unacceptable or questionable.

Fixes #41.


Conan-Kudo:
@ralphbean The only real issue I see is that you don't have something like how librepo does, where there's a `PYTHON_DESIRED` switch that can be set. For example, in the [librepo spec](http://pkgs.fedoraproject.org/cgit/librepo.git/tree/librepo.spec#n105), you can trigger it to search for Python 3 by passing in `-DPYTHON_DESIRED:str=3` to `%cmake`.

ralphbean:
Agreed @Conan-Kudo!  Should be a little nicer after e0b32cb.

ralphbean:
Any further comments here?  Anything else to fix?

Conan-Kudo:
@ralphbean I don't see anything.

@Tojaj Is there anything here?

Tojaj:
@ralphbean I finally did the review and the PR looks great. Thank you Ralph!

ralphbean:
Woot!  Thanks @Tojaj (and thanks @Conan-Kudo!)

PR #44, opened 2015-12-06
Replace on_exit() call with atexit() call (Fixes FTBFS on Unity Linux)

Conan-Kudo:
For [Unity Linux](http://unitylinux.com/index.php?id=about), we're using MUSL libc instead of glibc, so a great deal of non-standard and deprecated functions are not supported. This causes the createrepo_c build to fail because it uses `on_exit()`.

The `on_exit()` function is non-portable and is not available in all environments (notably Linux environments running on MUSL and Solaris based environments). 

Thus, the function call is swapped for the portable `atexit()` call.

The build and tests all pass after swapping `on_exit()` for `atexit()`.

### DIFF
--- a/src/createrepo_shared.c
+++ b/src/createrepo_shared.c
@@ -84,7 +84,7 @@ cr_set_cleanup_handler(const char *lock_dir,
         global_tmp_out_repo = NULL;
 
     // Register on exit cleanup function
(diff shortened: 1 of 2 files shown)


Tojaj:
Thanks!

PR #52, opened 2016-03-02
tags for old suse versions

M0ses:


Conan-Kudo:
@M0ses Could you guard this out with some kind of switch? I don't think it would make sense for this tag logic to be activated by default. Not to mention, you've explicitly noted this is for legacy stuff, so I would hesitate to suggest that it's something you'd want to have stick around forever.

Conan-Kudo:
Looks good to me :+1: 

@Tojaj: What do you think about it?

Tojaj:
@Conan-Kudo Agree!

@M0ses Thank you!

Tojaj:
Btw. I've just pushed two patches that update output printed when `--version` is used and when version is printed during verbose logging (`--verbose`).

Now there is a list of enabled compile-time features, e.g.:

    Version: 0.10.0 (Features: DeltaRPM LegacyWeakdeps )

M0ses:
cool, the features list will be a great help for debugging.

PR #54, opened 2016-03-24
Split option similar to python version of createrepo

M0ses:
* createrepo_c supports now the "--split" option
* base url is set to "media:" by default if no "-u" option is set
* trailing double slashes will be cut off ( "media://" -> "media:")

### DIFF
--- a/src/cmd_parser.c
+++ b/src/cmd_parser.c
@@ -106,6 +106,8 @@ static GOptionEntry cmd_entries[] =
       "Skip the stat() call on a --update, assumes if the filename is the same "
       "then the file is still the same (only use this if you're fairly "
       "trusting or gullible).", NULL },
+    { "split", 0, 0, G_OPTION_ARG_NONE, &(_cmd_options.split),
+      "generate split media", NULL },
     { "pkglist", 'i', 0, G_OPTION_ARG_FILENAME, &(_cmd_options.pkglist),
       "Specify a text file which contains the complete list of files to "
       "include in the repository from the set found in the directory. File "
--- a/src/createrepo_c.c
+++ b/src/createrepo_c.c
@@ -265,7 +265,7 @@ fill_pool(GThreadPool *pool,
         ++*package_count;
     }
 
-    return package_count;
+    return *package_count;
 }
 
 
@@ -818,7 +818,6 @@ main(int argc, char **argv)
     user_data.fil_db            = fil_db;
     user_data.oth_db            = oth_db;
     user_data.changelog_limit   = cmd_options->changelog_limit;
-    user_data.location_base     = cmd_options->location_base;
     user_data.checksum_type_str = cr_checksum_name_str(cmd_options->checksum_type);
     user_data.checksum_type     = cmd_options->checksum_type;
     user_data.checksum_cachedir = cmd_options->checksum_cachedir;
@@ -844,6 +843,11 @@ main(int argc, char **argv)
     user_data.deltatargetpackages = NULL;
     user_data.cut_dirs          = cmd_options->cut_dirs;
     user_data.location_prefix   = cmd_options->location_prefix;
+
+    if ( cmd_options->split && ! cmd_options->location_base)
+        user_data.location_base     = "media://";
+    else
+        user_data.location_base     = cmd_options->location_base;
 
     g_debug("Thread pool user data ready");
 
--- a/src/createrepo_c.c
+++ b/src/createrepo_c.c
@@ -346,7 +346,15 @@ main(int argc, char **argv)
         exit(EXIT_SUCCESS);
     }
 
-    if ( ! cmd_options->split ) {
+    if ( cmd_options->split ) {
+        if (argc < 2) {
+            g_printerr("Must specify at least one directory to index.\n");
+            g_printerr("Usage: %s [options] <directory_to_index> [directory_to_index] ...\n\n",
+                     cr_get_filename(argv[0]));
+            free_options(cmd_options);
+            exit(EXIT_FAILURE);
+        }
+    } else {
         if (argc != 2) {
             // No mandatory arguments
             g_printerr("Must specify exactly one directory to index.\n");
--- a/src/createrepo_c.c
+++ b/src/createrepo_c.c
@@ -147,7 +147,6 @@ fill_pool(GThreadPool *pool,
         GQueue *sub_dirs = g_queue_new();
         gchar *input_dir_stripped;
 
-
         input_dir_stripped = g_string_chunk_insert_len(sub_dirs_chunk,
                                                        in_dir,
                                                        in_dir_len-1);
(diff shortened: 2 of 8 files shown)


Conan-Kudo:
What is "split media" in the first place? Your description is somewhat lacking...

src/cmd_parser.c
```diff
@@ -106,6 +106,8 @@ static GOptionEntry cmd_entries[] =
       "Skip the stat() call on a --update, assumes if the filename is the same "
       "then the file is still the same (only use this if you're fairly "
       "trusting or gullible).", NULL },
+    { "split", 0, 0, G_OPTION_ARG_NONE, &(_cmd_options.split),
+      "generate split media", NULL },
```
Conan-Kudo: What is "split media"? This seems somewhat unhelpful.

M0ses:
Sorry , not meant as an excuse for lacking description, but this is the original description in createrepo.

What do you think about:

* generat split media (create meta data for distributed repositories)

or: (taken from manpage of createrepo)

* take a set of directories corresponding to different volumes in a media set and creates meta data in first directory

Conan-Kudo:
The latter description seems more helpful.

doc/createrepo_c.8
```diff
@@ -90,6 +90,10 @@ Optional output directory.
 .SS \-S \-\-skip\-symlinks
 .sp
 Ignore symlinks of packages.
+.SS \--split
+.sp
+Run in split media mode. Rather than pass a single directory, take a set of
+directories corresponding to different volumes in a media set.
```
Conan-Kudo: It's good that you added it to the man page, but I would prefer that this be also in the command parser code because occasionally the man pages are regenerated from the source code. See the utils directory for the scripts that can do this.

I prefer not having to regenerate the man pages, but I'd rather not run the risk of losing good descriptions due to someone running it.

M0ses: Thanks for your hint, fixed it.

Conan-Kudo:
Looks good to me :+1: 

@Tojaj Is there anything else that he needs to do? Or is it good to merge?

src/dumper_thread.c
```diff
@@ -230,7 +230,8 @@ load_rpm(const char *fullpath,
          int changelog_limit,
          struct stat *stat_buf,
          cr_HeaderReadingFlags hdrrflags,
-         GError **err)
+         GError **err,
```
Tojaj: Please keep code consistent and put GError as a last argument, this is an internal function anyway, you don't have to worry about API/ABI breakage

Tojaj:
@M0ses Thank you Frank, I quickly review the PR and it looks really good.
I just note a one minor thing about GError which is not listed as a last argument, could you please fix that?

src/createrepo_c.c
```diff
@@ -339,13 +346,15 @@ main(int argc, char **argv)
...
+    if ( ! cmd_options->split ) {
+        if (argc != 2) {
+            // No mandatory arguments
+            g_printerr("Must specify exactly one directory to index.\n");
+            g_printerr("Usage: %s [options] <directory_to_index>\n\n",
+                     cr_get_filename(argv[0]));
+            free_options(cmd_options);
+            exit(EXIT_FAILURE);
+        }
     }
```
Tojaj: Even when ``--split`` is specified, the number of arguments should still be checked, we need handle situation when there is no argument at all.

    $ ./createrepo_c --split
    Directory (null) must exist

is not right

Tojaj:
I wasn't able to test the code on Friday as I promised, sorry for that. But I tested it today and found two more issues.
* When ``--split`` is used, you skip check of number of arguments and thus it's able to run createrepo_c without any path specified.
* When ``--update`` is used, the base_url is not generated correctly ``./createrepo_c --verbose --update --split rpms_1/ rpms_3/`` (where repodata in rpms_1/ already exists) produces primary.xml where location base is set as ``<location xml:base="media://" ...``

M0ses:
> * When --split is used, you skip check of number of arguments and thus it's able to run createrepo_c without any path specified.

ok, I will change this for split option to require at least one directory - even it makes not much sense to use '--split' when you have only one directory/media, I think this is much more handy while scripting.

> When --update is used, the base_url is not generated correctly ./createrepo_c --verbose --update --split rpms_1/ rpms_3/ (where repodata in rpms_1/ already exists) produces primary.xml where location base is set as <location xml:base="media://" ...

Using the base_url from rpms_1 would be the expected behavior, right ?

M0ses:
Now I have seen the problem with the '--update' option and will try to fix it.
I hope the fix for checking the given input directories is ok for you.

Tojaj:
Repodata generated when ``--update`` is used should be identical with repodata generated without it.

src/dumper_thread.c
```diff
@@ -399,6 +430,13 @@ cr_dumper_thread(gpointer data, gpointer user_data)
         }
     } else {
         // Just gen XML from old loaded metadata
+        if ( task->media_id ) {
+            // need chunk to store location_base foreach package
+            if ( ! md->chunk )
+                md->chunk = g_string_chunk_new (PACKAGE_CHUNK_SIZE);
```
Tojaj: This isn't necessary and in fact it adds overhead because of thousands of new chunks are created, and each with size 2kB.
You should use the same approach as used for ``location_href`` in this function:
https://github.com/M0ses/createrepo_c/blob/split_option/src/dumper_thread.c#L345-L346
I.e. Change definition of ``location_base`` from ``const gchar *location_base`` into ``_cleanup_free_ gchar *location_base = NULL;``. Then you should malloc the location_base string and store it into this variable (note it will be automatically freed when program flow leaves the context of the function).
You only need to handle buffering in the same way as it's done for location_href - just use ``strdup`` for the location_base and add the copy to the buffered task:
https://github.com/M0ses/createrepo_c/blob/split_option/src/dumper_thread.c#L498-L499
Also don't forget to add a free call after this line:
https://github.com/M0ses/createrepo_c/blob/split_option/src/dumper_thread.c#L575

This has several benefits:
* You don't have to deal with GStringChunks which brings a lot of overhead if used for only one and short string.
* Handling of ``location_href`` and ``location_base`` will be consistent.
* You don't have to export ``PACKAGE_CHUNK_SIZE`` in ``package.h`` header file. (Btw. it's bad to export a constant without library prefix (``CR_``) in header file which is part of public API. Also this constant is pretty useless for external consumers and it should stay hidden)

Btw. you don't have to fix this, I can take a look on this tomorrow or maybe during Friday and do the needed refactoring.

M0ses:
@Tojaj Thanks a lot for your hints and explanations. Hopefully my new commit is ok for you, so you could accept the PR. Please let me know if you want me to stash the changes together to one single commit. I did not do so far to make it easier for you to review.

Tojaj:
Hi @M0ses 
I've just merged your commits, but I added a patch that refactors them a little: b8112ee97659184cf8469cfd97348b50e1a9823e

* I've removed redundant third param in ``prepare_split_media_baseurl``
* I've moved ``prepare_split_media_baseurl`` call from ``load_rpm`` and instead of call it from two different place (``prepare_split_media_baseurl`` and from ``cr_dumper_thread``) I call it only from ``cr_dumper_thread``.
* I've also refactored the ``prepare_split_media_baseurl`` to be more readable (it uses glib2 functions now).

Tojaj:
Btw. Thank you for your contribution!

M0ses:
@Tojaj  THX for refactor and merge

PR #59, opened 2016-04-22
added switch to turn of python support

M0ses:
Just some small modifications in the cmake files, so we can reduce the dependencies when building on older distributions


Conan-Kudo:
Looks good to me! 👍 

@Tojaj: Is there anything else that needs to be done?

Tojaj:
@M0ses Thank you!

Issue #58, opened 2016-04-21
Exit status should not be 0 on failure


nbartos:
When there is an issue opening a file (for example, a broken symlink), the error is reported, however the exit status still reports success.

The original createrepo returns failure if an error occurs opening a file.  While I really, really want to use createrepo_c because of the amazing speed, until this is fixed I have no choice but to use the older, slower, createrepo.

noelbk:
Here's a PR: https://github.com/rpm-software-management/createrepo_c/pull/63

PR #63, opened 2016-09-19
createrepo_c should exit nonzero if there are any errors.

noelbk:
Fixes Issue: https://github.com/rpm-software-management/createrepo_c/issues/58

This adds a flag to UserData for the cr_dumper_thread workers to record if theyt encountered an error, so main() can exit nonzero.

### DIFF
--- a/src/cmd_parser.c
+++ b/src/cmd_parser.c
@@ -186,6 +186,8 @@ static GOptionEntry cmd_entries[] =
       "Append this prefix before location_href in output repodata", "PREFIX" },
     { "repomd-checksum", 0, 0, G_OPTION_ARG_STRING, &(_cmd_options.repomd_checksum),
       "Checksum type to be used in repomd.xml", "CHECKSUM_TYPE"},
+    { "error-exit-val", 0, 0, G_OPTION_ARG_NONE, &(_cmd_options.error_exit_val),
+      "Exit with retval 2 if there were any errors during processing", NULL },
     { NULL, 0, 0, G_OPTION_ARG_NONE, NULL, NULL, NULL },
 };
 
--- a/src/cmd_parser.h
+++ b/src/cmd_parser.h
@@ -105,6 +105,7 @@ struct CmdOptions {
     gchar *location_prefix;     /*!< Append this prefix into location_href
                                      during repodata generation. */
     gchar *repomd_checksum;     /*!< Checksum type for entries in repomd.xml */
+    gboolean error_exit_val;        /*!< exit 2 on processing errors */
 
     /* Items filled by check_arguments() */
 
--- a/src/createrepo_c.c
+++ b/src/createrepo_c.c
@@ -864,8 +864,8 @@ main(int argc, char **argv)
     g_thread_pool_free(pool, FALSE, TRUE);
 
     // if there were any errors, exit nonzero
-    if( user_data.had_errors ) {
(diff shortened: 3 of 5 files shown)


Conan-Kudo:
It was specifically decided some time ago that the default behavior of createrepo_c was to not quit on errors in packages, but instead to warn so that the issues are known and most of the needed metadata ships.

However, this might be acceptable if a command-line switch was added to make it change behaviors as needed, preserving the current behavior as default, and the proposed behavior available as a switch.

noelbk:
This patch doesn't make createrepo_c exit early, it just exits nonzero if there was an error during the run.  So, createrepo_c will behave exactly the same as before, but it will return a different exit code on errors.  

@Conan-Kudo, do you still want a command-line switch to *always* exit 0?

Conan-Kudo:
@noelbk I misunderstood what this actually does. From my point of view, exiting with non-zero if there was issues is fine, as long as the repodata is generated. But I'm not sure how well tools that call to createrepo_c would handle it.

noelbk:
There's a chance that shell scripts that "set -e" could fail if createrepo_c exits with an error, but IMHO, it's better to be explicit about what errors you're ignoring, and "createrepo_c || true" would mask the nonzero exit.

noelbk:
@Conan-Kudo, it's been a while and I forgot about this!  What do you say, could we merge it?

Conan-Kudo:
@Tojaj What do you think of this?

Tojaj:
@noelbk Thank you for this PR! I like it, but I'm afraid that this is breaking "CLI API". Currently exit code ``1`` is returned when something serious happen and repodata aren't available. With this patch, you can have valid repodata but you don't have a way how to find that from a bash script because createrepo_c returned ``1``.

Thus it would be nice if the return code could be different for such non-fatal errors. e.g. ``2`` could work. With this we can document this change in behavior and tell people that when createrepo_c returns ``2`` it means that repodata are ok and available, but that there was something during generation which may need their attention.

Also I would  appreciated if there could be an option for disable or enable this behavior. Either something like ``--relaxed`` option that would cause return of ``0`` instead of ``2`` when non-fatal error happen or something like ``--strict``option that would enable the behavior of this patch.

Tojaj:
I noticed that this patch changes difference in behavior of ``createrepo`` and ``createrepo_c``:

Setup:

    # Create a repo
    mkdir repo
    # Add a package to the repo
    cp prelink-0.5.0-1.fc19.src.rpm repo/
    # Create a pkglist with one non-existing package
    echo -e "prelink-0.5.0-1.fc19.src.rpm\nfoobarpkg.rpm" > repo/pkglist.in

createrepo:

    createrepo --pkglist repo/pkglist.in repo/ 2>/dev/null 1>/dev/null; echo $?
    0

createrepo_c with the patch:

    ./createrepo_c --pkglist repo/pkglist.in repo/ 2>/dev/null 1>/dev/null; echo $?
    1

Which means that original ``createrepo`` actually doesn't mind when some packages from pgklist are missing, while ``createrepo_c`` with this patch is more strict about such "issues".

From what I know koji build system uses ``--pkglist`` directive quite often so maybe (not sure) it expects a behavior where createrepo is not failing on missing packages.
This is an additional argument for an option that could enable or disable behavior introduced by this patch.

Tojaj:
I've just tested the PR and it looks good.

@noelbk Thank you sir!

PR #64, opened 2016-09-26
Make set_record act like a setter

puiterwijk:
This will make sure that when set_record is called, all existing
records of the same type are removed.
It makes no sense to have multiple records of the same type,
and it actively breaks libhifs checksum validation.

Signed-off-by: Patrick Uiterwijk puiterwijk@redhat.com


puiterwijk:
Any chance this could be merged and picked up in a new release? Thanks!

Tojaj:
Merged, thanks!

PR #68, opened 2016-10-04
Ensure that provides/requires versions are factored into dep filtering

### LINKED ISSUE #67: createrepo_c filters requires from a package in metadata even when it provides a different version of it
It appears that createrepo_c is filtering Requires based on whether the package in question also Provides the same capability. However, the with that is it is not taking into account the version associated with the Provides when filtering Requires. Thus, situations where a package provides an older version of the capability but requires a newer version leads to the capability being stripped from the Requires, which breaks the install.

This has occurred with the `mutter` package with GNOME 3.22 in Mageia Cauldron (development for Mageia 6).

Reference: https://bugs.mageia.org/show_bug.cgi?id=19509

Conan-Kudo:
This PR includes a patch from @mlschroe that fixes #67.

Reference: https://bugs.mageia.org/show_bug.cgi?id=19509

### DIFF
--- a/src/parsehdr.c
+++ b/src/parsehdr.c
@@ -319,7 +319,10 @@ cr_package_from_header(Header hdr,
     };
 
     // Hastable with filenames from provided
-    GHashTable *provided_hashtable = g_hash_table_new(g_str_hash, g_str_equal);
+    GHashTable *provided_hashtable = g_hash_table_new_full(g_str_hash,
+                                                           g_str_equal,
+                                                           NULL,
+                                                           free);
 
     // Hashtable with already processed files from requires
     GHashTable *ap_hashtable = g_hash_table_new_full(g_str_hash,
@@ -358,15 +361,19 @@ cr_package_from_header(Header hdr,
                     }
 
                     // Skip package primary files
-                    if (g_hash_table_lookup_extended(filenames_hashtable, filename, NULL, NULL)) {
+                    if (*filename == '/' && g_hash_table_lookup_extended(filenames_hashtable, filename, NULL, NULL)) {
                         if (cr_is_primary(filename)) {
                             continue;
                         }
                     }
 
                     // Skip files which are provided
-                    if (g_hash_table_lookup_extended(provided_hashtable, filename, NULL, NULL)) {
-                        continue;
+                    gpointer pvalue;
+                    if (g_hash_table_lookup_extended(provided_hashtable, filename, NULL, &pvalue)) {
+                        struct ap_value_struct *ap_value = pvalue;
+                        if (!ap_value->flags || !ap_value->flags[0] || !flags || !flags[0] ||
+                            (!g_strcmp0(ap_value->flags, flags) && !strcmp(ap_value->version, full_version)))
+                           continue;
                     }
 
                     // Calculate pre value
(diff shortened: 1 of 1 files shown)


src/parsehdr.c
```diff
@@ -358,15 +361,19 @@ cr_package_from_header(Header hdr,
...
                     }
 
                     // Skip files which are provided
-                    if (g_hash_table_lookup_extended(provided_hashtable, filename, NULL, NULL)) {
-                        continue;
+                    gpointer pvalue;
+                    if (g_hash_table_lookup_extended(provided_hashtable, filename, NULL, &pvalue)) {
+                        struct ap_value_struct *ap_value = pvalue;
+                        if (!ap_value->flags || !ap_value->flags[0] || !flags || !flags[0] ||
+                            (!g_strcmp0(ap_value->flags, flags) && !strcmp(ap_value->version, full_version)))
```
ignatenkobrain: `strcmp` -> `g_strcmp0`

src/parsehdr.c
```diff
@@ -319,7 +319,10 @@ cr_package_from_header(Header hdr,
     };
 
     // Hastable with filenames from provided
-    GHashTable *provided_hashtable = g_hash_table_new(g_str_hash, g_str_equal);
+    GHashTable *provided_hashtable = g_hash_table_new_full(g_str_hash,
+                                                           g_str_equal,
+                                                           NULL,
+                                                           free);
```
ignatenkobrain: `g_free`

src/parsehdr.c
```diff
@@ -413,8 +420,12 @@ cr_package_from_header(Header hdr,
 
                 switch (deptype) {
                     case DEP_PROVIDES:
-                        g_hash_table_replace(provided_hashtable, dependency->name, dependency->name);
                         pkg->provides = g_slist_prepend(pkg->provides, dependency);
+                        struct ap_value_struct *pvalue = malloc(sizeof(struct ap_value_struct));
```
ignatenkobrain: `g_malloc`

src/parsehdr.c
Conan-Kudo: Addressed in glibify commit

src/parsehdr.c
Conan-Kudo: Addressed in glibify commit

src/parsehdr.c
mlschroe: (you're aware that all those non-glib code is just copy'n'paste from some lines above/below? Now doing some glibification makes the code pretty inconsistent...)

ignatenkobrain: @mlschroe then someone have to change code from where it was copied ;)

ignatenkobrain:
@Conan-Kudo also would be nice to get test for this.

Conan-Kudo:
@ignatenkobrain I'm not sure how to write tests for this.

Conan-Kudo:
@Tojaj What do you think of the fix?

src/parsehdr.c
```diff
@@ -358,15 +361,19 @@ cr_package_from_header(Header hdr,
...
                         }
                     }
 
                     // Skip files which are provided
-                    if (g_hash_table_lookup_extended(provided_hashtable, filename, NULL, NULL)) {
-                        continue;
+                    gpointer pvalue;
+                    if (g_hash_table_lookup_extended(provided_hashtable, filename, NULL, &pvalue)) {
+                        struct ap_value_struct *ap_value = pvalue;
+                        if (!ap_value->flags || !ap_value->flags[0] || !flags || !flags[0] ||
```
Tojaj: If I understand this correctly then:
I we already have a file which doesn't have any flags in the hash table (``!ap_value->flags``) then we don't care about any other files with the same filename even if flags and versions are different?

src/parsehdr.c
```diff
@@ -358,15 +361,19 @@ cr_package_from_header(Header hdr,
...
+                            (!g_strcmp0(ap_value->flags, flags) && !g_strcmp0(ap_value->version, full_version)))
+                           continue;
```
Tojaj: As the condition is on two lines and indentation could be confusing, could you please put curly brackets around the ``continue;`` single line scope?

Tojaj:
I created a new PR #70 which uses name-flags-version for dep checking, thus with this patch, if a package provides ``foo <= 20`` and requires ``foo > 20``then it should be processed correctly (the provide shouldn't be stripped).

Conan-Kudo:
Closing in favor of #70

PR #70, opened 2016-12-26
Consider versions during dep filtering for requires/provides (Issue #67)

Tojaj:
This should avoid situations where a package provides and requires same
capability (a dependency) with different versions but createrepo_c
strip the require down (because it already saw it in provides).

This should fix the issue in a better way than PR #68

### DIFF
--- a/src/parsehdr.c
+++ b/src/parsehdr.c
@@ -319,7 +319,10 @@ cr_package_from_header(Header hdr,
     };
 
     // Hastable with filenames from provided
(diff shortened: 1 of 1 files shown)


Tojaj:
Hi @Conan-Kudo, 
could you please try this patch (instead of patch in #68)?

Conan-Kudo:
@Tojaj Hey, sorry it took so long, but yes, I've verified that your fix does indeed solve [mga#19509](https://bugs.mageia.org/show_bug.cgi?id=19509). So it's okay with me!

PR #73, opened 2017-03-03
Preserve old metadata when --update is used and nothing has changed.

vathpela:
In the case where createrepo_c has been called with "--update", if the
repodata for every package we find is satisfied from the existing cache,
do not update the metadata files at all.

This means rsync doesn't have to evaluate the files for changes, and in
the case where '--unique-md-filenames --update' is used, it means rsync
won't have to transfer the newly renamed files.  This saves
approximately 8k of bandwidth (and many round trips) per rsync run in
the cases where nothing in the repo has changed, and also reduces the
time during a scheduled batch job where the repo is out of sync.

In the '--unique-md-filenames --update' case, it also means it won't
rename the files (actually representing only a newer generation of the
exact same data) if '--simple-md-filenames' was previously used on the
same repository.

Signed-off-by: Peter Jones <pjones@redhat.com>

### DIFF
--- a/src/createrepo_c.c
+++ b/src/createrepo_c.c
@@ -833,6 +833,7 @@ main(int argc, char **argv)
     user_data.skip_symlinks     = cmd_options->skip_symlinks;
     user_data.repodir_name_len  = strlen(in_dir);
     user_data.package_count     = package_count;
+    user_data.old_used          = 0;
     user_data.skip_stat         = cmd_options->skip_stat;
     user_data.old_metadata      = old_metadata;
     user_data.mutex_pri         = g_mutex_new();
@@ -869,6 +870,16 @@ main(int argc, char **argv)
     }
 
     g_message("Pool finished%s", (user_data.had_errors ? " with errors" : ""));
+
+    g_debug("Package count: %ld old_used: %ld", package_count, user_data.old_used);
+    if (user_data.old_used == package_count && cmd_options->update) {
+        g_debug("All packages were perfect cache hits, not updating repo.");
+        cr_remove_dir(tmp_out_repo, NULL);
+        if (g_strcmp0(lock_dir, tmp_out_repo))
+            // If lock_dir is not same as temporary repo dir then remove it
+            cr_remove_dir(lock_dir, NULL);
+        exit(EXIT_SUCCESS);
+    }
 
     cr_xml_dump_cleanup();
 
--- a/src/dumper_thread.c
+++ b/src/dumper_thread.c
@@ -387,11 +387,13 @@ cr_dumper_thread(gpointer data, gpointer user_data)
 
             if (udata->skip_stat) {
                 old_used = TRUE;
+                udata->old_used++;
             } else if (stat_buf.st_mtime == md->time_file
                        && stat_buf.st_size == md->size_package
                        && !strcmp(udata->checksum_type_str, md->checksum_type))
             {
                 old_used = TRUE;
+                udata->old_used++;
             } else {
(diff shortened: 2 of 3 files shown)


Conan-Kudo:
@vathpela Can you please fix the authorship of the patch? It's set to `rpm-build <rpm-build>` instead of `Peter Jones <pjones@redhat.com>`. A `git commit --amend --author="Peter Jones <pjones@redhat.com>"` should do the trick.

vathpela:
Sorry about that; updated now as 1a00f78.

Tojaj:
Looks nice, I'm just wondering if this couldn't cause some undesired side-effects.

For example in deployments where different pieces of repodata than packages may change between two runs (revision, repo tags: ``--distro``, ``--content``, ``--repo``, group file, location-prefix, ..). Or even compression, I can imagine that multiple repodata could be generated for one repo but with different compression (gz for legacy systems, xz for new ones) and rsynced/copied to different locations between the runs.

I also wonder how this behave with ``--update-md-path``? Cache hits from repos loaded from update md paths will cause incomplete metadata.

Last but not least, relative paths are not considered during cache lookups. If package location changes between runs, then update with your optimization will fall flat.

I really like the idea and understand your use case but I don't think it should be default behavior of ``--update``. I would rather see a new option for this behavior if possible?

vathpela:
Okay, I can make it a separate option (but probably not for a couple of weeks; deadline season is hitting pretty hard).  Would you like it to also want it to check those other options and refuse to operate with them?

vathpela:
> Last but not least, relative paths are not considered during cache lookups. If package location changes between runs, then update with your optimization will fall flat.

So commenting on this separately - that's true, but for *my* use case that's not much of a problem.  Basically I'm rebuilding the same package NVREs over and over while functionality/bugfixes/dependencies converge, so the file paths are not changing even though the contents are.

It's certainly possible to do a better version of this that works if e.g. the NVRE changes each time but very little else does, but it would require either identifying which things to copy from the old repodata up front or (quite a bit easier, but less efficient) comparing each generated repodata file with its ancestor and preserving the file metadata and name on the copy in the new metadata directory.  That would mean that package set changes which change some repodata files but not others would happen with many of the flags you mentioned, and could also easily work with "--unique-md-filename --update --try-not-to-mess-with-what-doesnt-change-so-rsync-doesnt-go-nuts", keeping the original unique file names, but doing updates in them.

But I didn't have the time to write that, because this is a yak I was shaving as the deadlines fast approach, entirely because the extra rsync time was taking more time in my work flow than avoiding it was.

dmach:
@vathpela I like the idea. If nothing changes, timestamps and repomd.xml shouldn't change either.
I found that the proposed patch doesn't honor additional args specified by the user, such as --no-database.

Reproducer:
in an empty directory, run:
$ createrepo_c .
$ ls repodata/*.sqlite*
# sqlite files are present as they are generated by default

$ createrepo_c . --update  --no-database
$ ls repodata/*.sqlite*
# sqlite files should not be present

The same applies on:
$ createrepo_c . --update --compress-format=xz
# the compress format doesn't change even if requested by the user

Maybe the change we're looking for is to preserve the original timestamps (in repomd.xml and also mtime of individual files) if output file checksum doesn't change.

Conan-Kudo:
@vathpela Are you still interested in this patch? If you rebase this and account for @dmach's feedback, I'd happily merge this.

Conan-Kudo:
@vathpela @dmach Does someone still care about fixing this up to land in `createrepo_c`?

Conan-Kudo:
Closed in favor of #287 by @kontura.

PR #78, opened 2017-07-20
Correctly install the shared library

kanavin:
Existing logic that guesses the suffix to "lib" breaks down
in cross-compilation environments - let's use Cmake's standard
variable instead that can be explicitly set from the build setup.


cgwalters:
Looks sane to me!

dmach:
Createrepo_c with this patch failed to build on RHEL 6 (cmake 2.8.12.2).
Let's keep the current CMakeLists.txt and change it after we drop RHEL 6 compatibility.

PR #83, opened 2017-12-14
Add libglnx GLib cleanup backports, use in a new place

cgwalters:
Now that the cleanup macros are upstream in GLib, let's follow
the path of e.g. ostree/rpm-ostree and many other projects that
use them.  I added the "backports" bits from libglnx:
https://git.gnome.org/browse/libglnx

I left in place the previous `cleanup.h`, but at some point we
should try to do a mass conversion.

To demonstrate use, I changed one place, but there is *tons* of code here that
could be cleaned up. Note also the use of "move semantics" there.

I'm writing this patch as it's likely I might start some work
on this project as part of the rpm-ostree jigdo ♲📦 effort:
https://github.com/projectatomic/rpm-ostree/issues/1081

### DIFF
--- a/src/cleanup.h
+++ b/src/cleanup.h
@@ -22,6 +22,8 @@
 
 /* This file was taken from libhif (https://github.com/hughsie/libhif) */
 
+#include "glnx-backport-autocleanups.h"
+
 #ifndef __CR_CLEANUP_H__
 #define __CR_CLEANUP_H__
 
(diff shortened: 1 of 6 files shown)


ignatenkobrain:
I wonder why not to just bump glib req

cgwalters:
> I wonder why not to just bump glib req

Offhand I would be not at all surprised if it was a requirement for this project to run on RHEL6.  And as the maintainer of glib2 in RHEL, let me be the first to say one would have to present serious justification to bump glib2 there, and this doesn't count 😉

cgwalters:
:telephone_receiver: - is there a maintainer for this project doing PR review?

jrohel:
@cgwalters Hi
1. I don't like bundling glib macros into createrepo_c. But I understand, it is hack for RHEL6. So OK.
2. src/glnx-backports.c is not listed in CMakeLists.txt. Is it what you want?

dmach:
Colin, I hope you don't mind me closing this PR.
We'd prefer not to bundle any glib macros in the code.
The problem eventually disappears once we drop RHEL 6 support.

PR #84, opened 2018-01-08
Correctly identify Requires(missingok) dependencies as legacy Recommends

Conan-Kudo:
Prior to RPM 4.12, RPM supported declaring `Recommends` using a hint on the `Requires` tag. The `Requires(missingok)` form was not broadly used, but historically in Mageia and Mandriva, `Suggests` aliased to this at build-time.

### DIFF
--- a/src/parsehdr.c
+++ b/src/parsehdr.c
@@ -33,6 +33,7 @@
 
 #ifdef ENABLE_LEGACY_WEAKDEPS
 #define RPMSENSE_STRONG (1 << 27)
+#define RPMSENSE_MISSINGOK (1 << 19)
 #endif
 
 typedef enum DepType_e {
@@ -436,6 +437,12 @@ cr_package_from_header(Header hdr,
                         pkg->obsoletes = g_slist_prepend(pkg->obsoletes, dependency);
                         break;
                     case DEP_REQUIRES:
+#ifdef ENABLE_LEGACY_WEAKDEPS
+                        if ( num_flags & RPMSENSE_MISSINGOK ) {
+                            pkg->recommends = g_slist_prepend(pkg->recommends, dependency);
+                            break;
+                        }
+#endif
                         dependency->pre = pre;
 
                         // XXX: libc.so filtering ////////////////////////////


ignatenkobrain:
Suggests or Recommends?

Conan-Kudo:
They behaved as `Recommends` but were called `Suggests`. They were the first implementation of weak dependencies in RPM. 

It was implemented as a hint for tie-breaking, apparently? https://github.com/rpm-software-management/rpm/commit/0d73d4699b47f839db61580687fc972f48dd42d6

However, Mandrake elected to co-opt it for supporting weak-installed things in `urpmi`.

ignatenkobrain:
so that should be Suggest then, no?

ignatenkobrain:
in any case, is there reason to fix this ancient thing? ;)

Conan-Kudo:
I'm dealing with packages that are using the flag, and having them be `Requires` instead of `Recommends` is annoying...

I can change it to map to Suggests, but `Requires(missingok)` meant to install it, but if it was unsatisfiable, it wasn't the end of the world (which sounds like `Recommends`). This is how urpmi behaved in Mageia < 5 (Mageia 5 had rpm 4.12, and we had a patch that translated it to `RPMTAG_OLDRECOMMENDS`).

berolinux:
Fixing it will also help OpenMandriva's move from rpm5 to rpm.org - we have a few packages relying on it (that will get fixed over time, but...)

src/parsehdr.c
```diff
@@ -436,6 +437,12 @@ cr_package_from_header(Header hdr,
                         pkg->obsoletes = g_slist_prepend(pkg->obsoletes, dependency);
                         break;
                     case DEP_REQUIRES:
+#ifdef ENABLE_LEGACY_WEAKDEPS
+                        if ( num_flags & RPMSENSE_MISSINGOK ) {
```
cgwalters: nit: The rest of this codebase doesn't use space ` ` inside conditionals.

src/parsehdr.c
```diff
@@ -33,6 +33,7 @@
 
 #ifdef ENABLE_LEGACY_WEAKDEPS
 #define RPMSENSE_STRONG (1 << 27)
+#define RPMSENSE_MISSINGOK (1 << 19)
```
cgwalters: Not a new issue, but why aren't we doing `#include <rpm/rpmds.h>`?  So we can build with older rpm?

src/parsehdr.c
Conan-Kudo: This logic was adapted from the one used with the `RPMSENSE_STRONG` check further down in the code.

src/parsehdr.c
Conan-Kudo: I imagine so. `createrepo_c` supports RHEL 6 and newer.

dmach:
The patch looks good to me, merging.

PR #92, opened 2018-06-12
Add zchunk support

jdieter:
This request adds zchunk support to createrepo_c.  It does now pass the python test cases, but there are some untested edge cases that will definitely need some work, namely using zchunk compression for anything other than primary, filelists and other.

The zchunk metadata is in addition to the standard gzip metadata and createrepo_c currently only creates zchunk versions primary, filelists and other.


Conan-Kudo:
@jdieter So this means that if I added other files to the metadata using `modifyrepo_c`, we can't chunk via zck for now?

jdieter:
That is correct.  I don't think it will be too hard to fix that, especially now that zchunk has the ability to chunk using the buzhash algorithm as well as user-specified blocks, but I first wanted to get the big three (well, more the big two: primary and filelists) sorted.

Conan-Kudo:
Well, as long as there is a plan for eventually supporting any arbitrary files manipulated by createrepo_c, this looks good to me.

src/cmd_parser.c
```diff
@@ -146,6 +146,14 @@ static GOptionEntry cmd_entries[] =
...
     { "xz", 0, 0, G_OPTION_ARG_NONE, &(_cmd_options.xz_compression),
       "Use xz for repodata compression.", NULL },
+    { "zck", 0, 0, G_OPTION_ARG_NONE, &(_cmd_options.zck_compression),
+      "Generate zchunk files as well as the standard repodata.", NULL },
+    { "zck-primary-dict", 0, 0, G_OPTION_ARG_FILENAME, &(_cmd_options.zck_primary_dict),
+      "Compression dictionary to use for zchunk primary file", "ZCK_PRIMARY_DICT" },
+    { "zck-filelists-dict", 0, 0, G_OPTION_ARG_FILENAME, &(_cmd_options.zck_filelists_dict),
+      "Compression dictionary to use for zchunk filelists file", "ZCK_FILELISTS_DICT" },
+    { "zck-other-dict", 0, 0, G_OPTION_ARG_FILENAME, &(_cmd_options.zck_other_dict),
+      "Compression dictionary to use for zchunk other file", "ZCK_OTHER_DICT" },
```
Conan-Kudo: Are the dictionaries mandatory, or optional? My understand of zstd is that they make it better for compression, but aren't necessary.

jdieter: They are optional, but make a significant difference in the size of the zchunked metadata.  I've put together dictionaries for primary, filelists and other based on data from current Fedora releases, that I'd suggest packaging for Fedora/EPEL (if that's what our builders are on).

The dictionary is embedded in the zchunk file, so clients don't need a copy.

If the dictionary changes, there will be no common chunks between the pre-change and post-change files, so dictionary changes should happen a maximum of once a release (and probably aren't needed even that often)

Conan-Kudo: Forgive me for being forgetful, but is there a script or something that people can use to generate dictionaries for their distributions' metadata?

jdieter: I've got a [small python script](https://www.jdieter.net/downloads/zchunk-dicts/split.py) that splits the metadata into much smaller files and strips out any checksums (which by definition are unpredictable).

I ran the script over a couple of days worth of updates metadata, which produced thousands of files, and then, in the directory containing those files, ran:
```
zstd --train --dictID=0 *
```

The default dict size of roughly 100KB was fairly optimal for Fedora's metadata.  Too small and you don't see the savings you'd expect.  Too large and you're just wasting space.

jdieter: The proposed Fedora dicts are available at [https://www.jdieter.net/downloads/zchunk-dicts](https://www.jdieter.net/downloads/zchunk-dicts)

jdieter: I've gone ahead and updated the man page.  I'll work on the tests but it might take a bit longer to get them done.

src/misc.c
```diff
@@ -1256,6 +1256,46 @@ cr_write_to_file(GError **err, gchar *filename, const char *format, ...)
     return ret;
 }
 
+gchar *cr_read_from_file(GError **err, gchar *filename, size_t *file_size) {
```
cgwalters: How is this different from `g_file_get_contents()`?

jdieter: Wow, I feel stupid.  It's exactly the same.  I've just pushed a commit that remove cr_read_from_file and uses g_file_get_contents instead.

jdieter:
@Conan-Kudo, I've added some tests (and found and fixed some problems in the process).  Do you have some further suggestions as to which tests I should create?

jdieter:
Just a heads up that the latest tests require zchunk-0.7.3 (available in the copr) in order to work, as it added the --stdout flag to unzck.

Conan-Kudo:
No, it looks good to me.

CMakeLists.txt
```diff
@@ -34,7 +34,13 @@ find_package(LZMA REQUIRED)
 find_package(OpenSSL REQUIRED)
 find_package(Sqlite3 REQUIRED)
 find_package(ZLIB REQUIRED)
-
+find_package(PkgConfig REQUIRED)
+find_library(ZCHUNKLIB NAMES zck)
```
Conan-Kudo: Could you write a `FindZChunk.cmake` similar to the `FindGLIB2.cmake` module and use that instead?

jdieter: Ok, let me know if this works or you want something different

src/createrepo_c.c
```diff
@@ -818,13 +819,165 @@ main(int argc, char **argv)
...
+                                            &tmp_err);
+        assert(oth_cr_zck || tmp_err);
+        if (!oth_cr_zck) {
+            g_critical("Cannot open file %s: %s",
+                       oth_zck_filename, tmp_err->message);
+            g_clear_error(&tmp_err);
+            cr_contentstat_free(pri_zck_stat, NULL);
+            cr_contentstat_free(fil_zck_stat, NULL);
+            cr_contentstat_free(oth_zck_stat, NULL);
+            g_free(pri_zck_filename);
```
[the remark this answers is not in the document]
jdieter: g_autofree() was added in glib-2.44.  What's the minimum glib version that we need to support?

CMakeLists.txt
```diff
@@ -34,6 +34,8 @@ find_package(LZMA REQUIRED)
 find_package(OpenSSL REQUIRED)
 find_package(Sqlite3 REQUIRED)
 find_package(ZLIB REQUIRED)
+find_package(PkgConfig REQUIRED)
+find_package(Zchunk REQUIRED)
```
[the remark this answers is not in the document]
jdieter: I didn't know about pkg_check_modules().  Thanks for the pointer!

jdieter:
There are no zck-specific include directories.  zck.h is in /usr/include.  Should I still have include_directories(...)?

Conan-Kudo:
@jdieter Yes, because it might be installed to non-standard paths by someone else.

jdieter:
Ok, is this right?

Conan-Kudo:
Looks good to me.

j-mracek:
Please whats a status of availability packages that fulfill requirement "pkg_check_modules(ZCK REQUIRED zck)" in Fedora/Centos/RHEL ... ?

Conan-Kudo:
@j-mracek It's been available in Fedora, CentOS, RHEL, openSUSE, and Mageia for a couple of weeks now.

j-mracek:
Please and whats a package name, because I was unable to find it?

Conan-Kudo:
`zchunk-devel` is the package you need installed for this code. That package name will work across all distributions (though this will install `lib64zck-devel` on Mageia, and `libzck-devel` on openSUSE).

jdieter:
I've been thinking about how zchunk-enabled createrepo_c uses dictionaries, and the current method is a bit awkward.  To use createrepo_c without dictionaries, you just run:
```
createrepo_c --zck ./
```

But to enable dictionary use, you need to run:
```
createrepo_c --zck --zck-primary-dict=foo-primary.dict --zck-filelists-dict=foo-filelists.dict --zck-other-dict=foo-other.dict ./
```

As we look at adding zchunk support to prestodelta.xml and other files generated by createrepo_c or added using modifyrepo_c, this starts to become very unwieldy.  I suggest a different solution, passing a ```--zck-dict-dir``` argument that points to a directory containing ```foo.dict``` where ```foo``` is ```primary```, ```filelists```, ```other```, ```prestodelta```, etc.  The command usage would look like:
```
createrepo_c --zck --zck-dict-dir=/usr/share/repo-dict
```

If there is no dictionary in the directory that matches the current file, then the file would be compressed without a dictionary.

Thoughts?  Should I go ahead and implement this as part of this patch set?

Conan-Kudo:
@jdieter That sounds fine to me, but I suggest that you document what it expects for that.

Conan-Kudo:
Rather than having that structure exclusively, it's probably better to support both ways.

So, for now, I'd like to see this go in as-is, and we can add the `--zck-dict-dir` option later.

jdieter:
@Conan-Kudo, if you feel strongly about it, we can keep the current options, but I feel like they're really bulky.  I think ```--zck-dict-dir``` is a more elegant method, and I'd definitely prefer to drop the current options before people start depending on them.

dmach:
I wrote couple comments and made a PR for the librepo changes.
A patch to make zchunk optional is required (similar to whe one I posted for librepo).
I can eventually make it, but I'll need a review.

jdieter:
I'll see if I can make some time to work on this and the librepo PR tomorrow morning.

jdieter:
@dmach Ok, I've gone ahead and written a patch to make zchunk optional, but it will currently fail the python tests when you compile with -DWITH_ZCHUNK=no.  I'm going to work on the tests next.

FWIW, I've been focusing on getting createrepo_c fully finished before fixing up the client side because we've changed how the zchunk metadata is stored in repomd.xml due to mailing list feedback.  The zchunk library has also had some API changes that I haven't yet ported to librepo.

If you'd prefer to review all of the patches simultaneously, we can do that, but my preference would be to get createrepo done first, and then do the others together afterwards.

jdieter:
@Conan-Kudo, I've gone ahead and implemented https://github.com/rpm-software-management/createrepo_c/pull/92#issuecomment-415347810.  When you set --zck-dict-dir, createrepo_c will look for a file in the directory that matches the metadata file and a .dict extension.  So for primary.xml, the directory would need to have primary.zck.dict.  This way, you can (eventually) use the same arguments when adding files using modifyrepo.

Conan-Kudo:
@jdieter So this does not yet work for `mergerepo_c` (used by Koji) and `modifyrepo_c` (used by COPR)?

jdieter:
@Conan-Kudo, you still can't run --zck in mergerepo_c or modifyrepo_c, but, at least as far as I've tested (which hasn't been extensive), they will keep any zchunk metadata that's already been added.

jdieter:
> @jdieter So this does not yet work for `mergerepo_c` (used by Koji) and `modifyrepo_c` (used by COPR)?

https://github.com/rpm-software-management/createrepo_c/pull/92/commits/cb111c3a1f1862d10f52c473d7b0a78fe5a03565 implements zchunk support in mergerepo_c, just for primary, filelists and other.

jdieter:
Is there anything else that needs to be done with this before it's reviewed?

jdieter:
@dmach, now that F29's out the door, I'd love to start generating zchunk metadata in official Fedora.  Is there anything I can do to help get this reviewed?

Conan-Kudo:
@jdieter Can you please rebase this change? There are reported conflicts from the other merge earlier today.

jdieter:
Done!  Thanks for the heads up!

Conan-Kudo:
@dmach Can we please merge this now?

src/compression_wrapper.c
```diff
@@ -26,9 +26,11 @@
 #include <string.h>
 #include <stdio.h>
 #include <ctype.h>
+#include <stdbool.h>
 #include <zlib.h>
 #include <bzlib.h>
 #include <lzma.h>
+#include <zck.h>
```
dmach: This needs to be wrapped into #ifdef WITH_ZCHUNK

jdieter: Ok, this is done now.  I did a test build on a system without zchunk, and it built and ran just fine.

jdieter: Thanks so much for committing these patches!!!

dmach:
I'll submit a patch to Fedora's dist-git, spec needs couple tweaks to handle bcond_with(out) zchunk

PR #96, opened 2018-06-21
Add the capability to build Python (PyPI) packages directly.

dralley:
* Added a setup.py and pyproject.toml
* Made CMake config file changes to facilitate building in a
  different context
* Updated gitignore

This will allow Python bindings (not the executables) for createrepo_c to be installed cross-distro.

To create a source distribution (requires the installer to have build dependencies installed), use ```python3 setup.py sdist```

To create a binary distribution (pre-compiled, but requires a similar platform), use ```python3 setup.py bdist_wheel```


src/python/CMakeLists.txt
```diff
@@ -69,5 +69,10 @@ TARGET_LINK_LIBRARIES(_createrepo_c
                         ${PYTHON_LIBRARY}
                      )
 
-INSTALL(FILES __init__.py DESTINATION ${PYTHON_INSTALL_DIR}/createrepo_c)
-INSTALL(TARGETS _createrepo_c LIBRARY DESTINATION ${PYTHON_INSTALL_DIR}/createrepo_c)
+IF (SKBUILD)
+    INSTALL(FILES __init__.py DESTINATION src)
+    INSTALL(TARGETS _createrepo_c LIBRARY DESTINATION src)
```
dralley: This bit is not ideal.  I ran into issues using a different directory, and I'm not sure yet whether it's because I was doing something wrong or if it's a bug w/ scikit-build.  I'll keep investigating.

It's not really a huge issue either, though.

dralley:
I thought of some ways to improve this, I'm going to close and re-open when ready.

CMakeLists.txt
```diff
@@ -126,20 +126,26 @@ endif()
...
-        execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink createrepo_c \$ENV{DESTDIR}${BASHCOMP_DIR}/modifyrepo_c)
-        execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink createrepo_c \$ENV{DESTDIR}${BASHCOMP_DIR}/sqliterepo_c)
-        ")
-ELSE (BASHCOMP_FOUND)
-    INSTALL(FILES createrepo_c.bash DESTINATION "/etc/bash_completion.d")
-    message("Bash completion directory: /etc/bash_completion.d")
-endif (BASHCOMP_FOUND)
+OPTION(ENABLE_BASHCOMP "Install Bash autocompletions?" ON)
+IF (SKBUILD)
+    SET (ENABLE_BASHCOMP OFF)
```
dralley: Python packages can't carry bash completions, so we need to disable them when building a Python package.

We're using Scikit-build to handle building the Python package, which defines SKBUILD when it's being used.

doc/CMakeLists.txt
```diff
@@ -8,9 +8,11 @@ if(DOXYGEN_FOUND)
         COMMENT "Building C API documentation with Doxygen" VERBATIM)
 endif(DOXYGEN_FOUND)
 
+IF (NOT SKBUILD)
 INSTALL(FILES createrepo_c.8 mergerepo_c.8 modifyrepo_c.8 sqliterepo_c.8
         DESTINATION share/man/man8
         COMPONENT bin)
+ENDIF (NOT SKBUILD)
```
dralley: Same with manual files ^

src/CMakeLists.txt
```diff
@@ -119,7 +121,8 @@ INSTALL(TARGETS createrepo_c DESTINATION bin/)
 INSTALL(TARGETS mergerepo_c DESTINATION bin/)
 INSTALL(TARGETS modifyrepo_c DESTINATION bin/)
 INSTALL(TARGETS sqliterepo_c DESTINATION bin/)
+ENDIF (NOT SKBUILD)
```
dralley: I think we the binaries could be included with the Python package *in theory*, but when I tried it it couldn't find the shared library.  No need to worry about it for now.

src/python/CMakeLists.txt
```diff
@@ -69,5 +69,10 @@ TARGET_LINK_LIBRARIES(_createrepo_c
                         ${PYTHON_LIBRARY}
                      )
 
-INSTALL(FILES __init__.py DESTINATION ${PYTHON_INSTALL_DIR}/createrepo_c)
-INSTALL(TARGETS _createrepo_c LIBRARY DESTINATION ${PYTHON_INSTALL_DIR}/createrepo_c)
+IF (SKBUILD)
```
dralley: Python package management will handle putting them in the right places for us.

PR #101, opened 2018-07-26
Add mergerepo_c --repo-prefix-search and --repo-prefix-replace.

hanzz:
I have service which merges multiple repositories together. The repositories live on NFS which is mounted in `/mnt/odcs/repos` in backend. The final merged repositories are stored in `/mnt/odcs/public`. I therefore call `mergerepo_c -o /mnt/odcs/public -r /mnt/odcs/repos/foo -r /mnt/odcs/repos/bar -r ...`.

The issue is that merged repository should be accessed using HTTP on certain domain (for example http://odcs.fedoraproject.org/composes/), but the merged metadata contains location base set to `file:///mnt/odcs/public`, so the merged repo is not useful.

This PR adds new `--repo-prefix` and `--repo-prefix-url` options to address this. In my case, I can use `--repo-prefix /mnt/odcs/public --repo-prefix-url http://odcs.fedoraproject.org/composes` and the `/mnt/odcs/public` in repopath is replaced by the `http://odcs.fedoraproject.org/composes` and the repodata is useful.

### DIFF
--- a/src/mergerepo_c.c
+++ b/src/mergerepo_c.c
@@ -115,6 +115,8 @@ struct CmdOptions {
 
     gboolean version;
     char **repos;
+    char *repo_prefix_search;
+    char *repo_prefix_replace;
     char *archlist;
     gboolean database;
     gboolean no_database;
@@ -164,6 +166,10 @@ static GOptionEntry cmd_entries[] =
       "Show program's version number and exit", NULL },
     { "repo", 'r', 0, G_OPTION_ARG_FILENAME_ARRAY, &(_cmd_options.repos),
       "Repo url", "REPOS" },
+    { "repo-prefix-search", 0, 0, G_OPTION_ARG_STRING, &(_cmd_options.repo_prefix_search),
+      "Repository prefix to be replaced by repo-prefix-url." },
+    { "repo-prefix-replace", 0, 0, G_OPTION_ARG_STRING, &(_cmd_options.repo_prefix_replace),
+      "Repository prefix URL by which the repo-prefix is replaced." },
     { "archlist", 'a', 0, G_OPTION_ARG_STRING, &(_cmd_options.archlist),
       "Defaults to all arches - otherwise specify arches", "ARCHLIST" },
     { "database", 'd', 0, G_OPTION_ARG_NONE, &(_cmd_options.database),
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
@hanzz This looks good, but could you please add a test case for this?

hanzz:
Hm, are there already some tests for mergerepo_c? I don't see anything in "tests". I could add tests for this single feature, but I think writing whole new test-suite for the mergerepo_c should not happen in this PR.

hanzz:
@dmach: Can we merge this? It is important fix for ODCS/OSBS team.

src/mergerepo_c.c
```diff
@@ -115,6 +115,8 @@ struct CmdOptions {
 
     gboolean version;
     char **repos;
+    char *repo_prefix;
+    char *repo_prefix_url;
```
dmach: Could you replace these (and also the options) to followin?:
* repo_prefix -> repo_prefix_search
* repo_prefix_url -> repo_prefix_replace

src/mergerepo_c.c
```diff
@@ -164,6 +166,10 @@ static GOptionEntry cmd_entries[] =
       "Show program's version number and exit", NULL },
     { "repo", 'r', 0, G_OPTION_ARG_FILENAME_ARRAY, &(_cmd_options.repos),
       "Repo url", "REPOS" },
+    { "repo-prefix", 0, 0, G_OPTION_ARG_STRING, &(_cmd_options.repo_prefix),
+      "Repository prefix to be replaced by repo-prefix-url." },
+    { "repo-prefix-url", 0, 0, G_OPTION_ARG_STRING, &(_cmd_options.repo_prefix_url),
+      "Repository prefix URL by which the repo-prefix is replaced." },
```
dmach: Make sure the options must be used together.

python-like meta-code follows :) :
if repo_prefix_search != NULL or repo_prefix_replace != NULL:
   (I assume that NULL means that option hasn't been set)
   if repo_prefix_search == NULL: error
   if repo_prefix_replace == NULL: error
   (they have to be used together)

   for repo in repos:
      search for prefix and replace it
      if repo_prefix_search == "": prepend new prefix to all repos
      if repo_prefix_replace == "": remove prefixes

(I hope the option parser allows this behavior)

hanzz:
@dmach: I've updated the PR to address your comments.

PR #103, opened 2018-07-30
Enable building Python packages for PyPI distribution

dralley:
This will allow Python bindings *and executables* for createrepo_c to be distributed cross-distribution via PyPI.

To create a source distribution (requires the installer to have build dependencies installed, but not the builder), use ```python3 setup.py sdist```.  

To create a binary distribution (pre-compiled, does not require the installer to have build deps, caveat [0]), use ```python3 setup.py bdist_wheel```.   Note: you do need to have ```scikit-build``` package installed to build the package.

To install the package, use ```pip install dist/createrepo_c-*.tar.gz``` if it's an sdist package or ```pip install dist/createrepo_c-*.whl``` if it's a bdist.  If it is an sdist package, the installer also needs to have the ```scikit-build``` package installed.  New versions of ```pip``` >= 10.0 should handle this for you due to the ```pyproject.toml``` file, but ```pip``` < 10.0 will not.

Once uploaded to PyPI the package can be installed with ```pip install createrepo-c```

Cmake build system changes
-------------------------------------

* Added a toggle ```ENABLE_BASHCOMP```, default ON, so that installing bash completions can be disabled if necessary.
* Added a toggle ```CREATEREPO_C_INSTALL_DEVELOPMENT```, default ON, so that installing header files and pkginfo can be disabled if necessary.
* Added a toggle ```CREATEREPO_C_INSTALL_MANPAGES```, default ON, so that installing manual pages can be disabled if necessary.
* Added a toggle ```BUILD_LIBCREATEREPO_C_SHARED```, default ON, so that libcreaterepo_c can be built statically if needed, e.g. built directly into the python extension library.
* Added ```OPENSSL_INCLUDE_DIR``` variable to enable building against local/custom OpenSSL libraries.

Since these are general changes, if you would like me to split these changes out into separate Pull Requests, I can do so.  However, they are prerequisites for this PR.  

Python specific changes
-------------------------------

* Added a setup.py, requirements-dev.txt and pyproject.toml
* Updated gitignore
* CMake will take different paths if ```SKBUILD``` is set.  ```SKBUILD``` is set by in the context of building a Python package by "scikit-build", which acts as a bridge between CMake and the Python packaging infrastructure.

[0] Binary distributions are not generally portable across different systems - a bdist created on Fedora 28 probably will not work on Ubuntu 16.04.  However, there is a specific environment provided by Python Packaging Authority called [manylinux](https://github.com/pypa/manylinux), from which you can produce widely portable bdists.  The current "manylinux1" environment is based on CentOS 5, which uses a GLib version too old to build createrepo_c.  A [new environment](https://github.com/pypa/manylinux/pull/182) is in the works based on CentOS 6, but it is not officially ready yet.  Thus, for the time being (but hopefully not for very long), it won't be possible to create a widely portable bdist package of createrepo_c.


dralley:
I'd like to give enormous thanks to @jcfr for his help with this PR.

jcfr:
Follow-up to this PR can be bound here: https://github.com/dralley/createrepo_c/pull/2

setup.py
```diff
@@ -0,0 +1,26 @@
...
+    url='https://github.com/rpm-software-management',
+    classifiers=(
+        'License :: OSI Approved :: GNU General Public License v2 (GPLv2)',
+        'Operating System :: POSIX :: Linux',
+        'Programming Language :: C',
+        'Topic :: System :: Software Distribution',
+        'Topic :: System :: Systems Administration',
+        'Programming Language :: Python :: 3',
+    ),
+    python_requires=">=3.5",
```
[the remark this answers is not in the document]
jcfr: Fixed in https://github.com/dralley/createrepo_c/pull/2/commits/ee6df5f2dc47fc7cf6d103fcea5041ab8d73fad4

setup.py
```diff
@@ -0,0 +1,26 @@
...
+    packages=['createrepo_c'],
+    package_dir={
+        'createrepo_c': 'src/python'
+    },
+    cmake_args=['-DPYTHON_DESIRED:STRING=3', '-DBUILD_LIBCREATEREPO_C_SHARED:BOOL=OFF'],
```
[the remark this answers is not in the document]
jcfr: addressed in https://github.com/dralley/createrepo_c/pull/2/commits/ee6df5f2dc47fc7cf6d103fcea5041ab8d73fad4

dralley:
Follow-up merged

setup.py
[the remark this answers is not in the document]
dralley: My thinking was that - what if you've built an sdist, and you want to install it with pip for Python 3, and the current CMake default is Python 2.  Or vice-versa, the default is Python 3, but you want to install an sdist for Python 2.  I don't think we're handling that situation gracefully.

Given that Python 2 only has a year and a half before EOL, I was thinking the benefit was not worth the cost of the added complexity.

CMakeLists.txt
```diff
@@ -126,20 +135,23 @@ endif()
...
-    INSTALL(CODE "
-        execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink createrepo_c \$ENV{DESTDIR}${BASHCOMP_DIR}/mergerepo_c)
-        execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink createrepo_c \$ENV{DESTDIR}${BASHCOMP_DIR}/modifyrepo_c)
-        execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink createrepo_c \$ENV{DESTDIR}${BASHCOMP_DIR}/sqliterepo_c)
-        ")
-ELSE (BASHCOMP_FOUND)
-    INSTALL(FILES createrepo_c.bash DESTINATION "/etc/bash_completion.d")
-    message("Bash completion directory: /etc/bash_completion.d")
-endif (BASHCOMP_FOUND)
+OPTION(ENABLE_BASHCOMP "Install Bash autocompletions?" ON)
```
dralley: Since we're being verbose for the other options, should this be named ```CREATEREPO_C_INSTALL_BASHCOMP``` to be more consistent?

jcfr: Yes. Ideally, option specific to the project should be prefixed by the project name.

jcfr: But this could be done in a different PR along with some other stylistic changes

src/python/__init__.py
```diff
@@ -326,3 +330,36 @@ def decompress_file(src, dst, comtype, stat=None):
 compression_suffix  = _createrepo_c.compression_suffix
 detect_compression  = _createrepo_c.detect_compression
 compression_type    = _createrepo_c.compression_type
+
+
+DATA = os.path.join(os.path.dirname(__file__), 'data')
+
+# Support running tests from the source tree
+if not os.path.exists(DATA):
+    from skbuild.constants import CMAKE_INSTALL_DIR as SKBUILD_CMAKE_INSTALL_DIR
```
dralley: We can't rely on this - most users won't be installing via PyPI, they'll be using the normal RPM.  And even if they're installing from PyPI, hopefully we'll be able to provide a bdist eventually, so there's no need to have a runtime dep on ```scikit-build```.

This causes an import error when following the normal build / test instructions in the README.

Putting that aside, running tests using the normal instructions has been broken by a change somewhere.  I get 

```
ERROR: Failure: ImportError (cannot import name _createrepo_c)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/usr/lib/python2.7/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/usr/lib/python2.7/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/home/vagrant/devel/createrepo_c/tests/python/tests/test_xml_parser.py", line 6, in <module>
    import createrepo_c as cr
  File "/home/vagrant/devel/createrepo_c/build/src/python/createrepo_c/__init__.py", line 8, in <module>
    from . import _createrepo_c
ImportError: cannot import name _createrepo_c
```

From running ``` PYTHONPATH=`readlink -f ./build/src/python/` nosetests -s tests/python/tests/````

despite that the path does have ```_createrepo_c.so``` on it.

```
[vagrant@pulp3 createrepo_c]$ ls build/src/python/createrepo_c/
_createrepo_c.so  __init__.py  __init__.pyc
```

Looking into what might be causing this currently.

src/python/CMakeLists.txt
```diff
@@ -28,6 +30,14 @@ set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fno-strict-aliasing")
 set (CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -fno-strict-aliasing")
 set (CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} -fno-strict-aliasing")
 
+IF (NOT SKBUILD)
+    IF (${CMAKE_VERSION} VERSION_LESS "2.8.0")
```
jcfr: Would it be sensible to bump the minimum required version fo CMake ?

dralley: @Conan-Kudo

src/python/__init__.py
jcfr: >We can't rely on this - most users won't be installing via PyPI

This was only to support the case of local installation  where the developer would create an editable install using `python install -e .` or `python setup.py develop`

> running tests using the normal build process documented in the README has been broken by a change somewhere

More details in this post: https://groups.google.com/d/msg/scikit-build/XYfZFT2iFJM/lFEaEnm3CQAJ

Conan-Kudo:
@dralley We can probably bump up to CMake 2.8.12.

src/python/__init__.py
jcfr: To summarize, there are two issues preventing editable from working.

By editable install, I mean the ability to run `pip install -e .` and then import `createrepo_c`

(1) python source should be moved from `src/python/*` to `src/python/createrepo_c/*`

(2) resolving the location of the compiled module should be fixed by either having a fallback to `skbuild.constant` or by hardcoding the expected location

Conan-Kudo: @dralley The python bindings are shipped in Fedora/CentOS, Mageia, and openSUSE. Only SUSE Linux Enterprise currently does not ship them.

dralley: @Conan-Kudo Ah, I hadn't noticed it was built as a separate RPM package (```python3-createrepo_c```).  Thanks.

dralley: @jcfr I've moved ```src/python/__init__.py``` to ```src/python/createrepo_c/__init__.py```, so I think everything is fine now on that front.  My editable installation worked.

re: 2, Here is a possible solution (incomplete commit, going to lunch)

https://github.com/rpm-software-management/createrepo_c/pull/103/commits/58637bf102d6345f83702e8c72297bb2db66a339

src/python/createrepo_c/__init__.py
```diff
@@ -326,3 +330,48 @@ def decompress_file(src, dst, comtype, stat=None):
...
+    # Try building the path manually.
+    try:
+        from skbuild.constants import CMAKE_INSTALL_DIR as SKBUILD_CMAKE_INSTALL_DIR
+        _data = os.path.abspath(os.path.join(
+            os.path.dirname(__file__), '..', SKBUILD_CMAKE_INSTALL_DIR, 'src/python/createrepo_c/data'))
+        if os.path.exists(_data):
+           _BIN_DIR = os.path.join(_data, 'bin')
+    except ImportError:
+        # Scikit isn't available
+        pass
```
jcfr: Need also to add 
```
else:
    _BIN_DIR = os.path.join(_DATA_DIR, 'bin')
```

Otherwise trying to run any of the executable after installing the wheel ends up in an infinite loop.

src/python/createrepo_c/__init__.py
```diff
@@ -326,3 +330,48 @@ def decompress_file(src, dst, comtype, stat=None):
...
+
+# Where we will look for the binaries. Default to looking on the system PATH.
+# If one of the following test succeeds, we can change this to look somewhere else.
+# Otherwise, probably, we were not built as a Python package, e.g. RPM, "cmake ..; make"
+# In that case, let's just assume that the binary will be on the PATH somewhere.
+_BIN_DIR = ""
+
+# Support running tests from the source tree
+if not os.path.exists(_DATA_DIR):
+    # We can't find it. This might mean we've been built as an "editable" Python installation
```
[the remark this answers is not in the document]
dralley: Didn't know about that technical detail, good point.

src/python/createrepo_c/__init__.py
```diff
@@ -332,21 +332,34 @@ def decompress_file(src, dst, comtype, stat=None):
...
+if os.path.exists(_DATA_DIR):
+    _BIN_DIR = os.path.join(_DATA_DIR, 'bin')
+else:
+    # We can't find it. This could mean we've been built as an "editable" Python installation.
+    # Try building the path manually if scikit-build is available. If it isn't, or if the path
+    # doesn't exist in the source tree, continue assuming it's just on the PATH.
+    try:
+        from skbuild.constants import CMAKE_INSTALL_DIR as SKBUILD_CMAKE_INSTALL_DIR
+        _data = os.path.abspath(os.path.join(
+            os.path.dirname(__file__), '../../..', SKBUILD_CMAKE_INSTALL_DIR, 'src/python/createrepo_c/data'))
```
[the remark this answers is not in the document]
jcfr: The equivalent of 

```
pip install -e .
```

would be 

```
python setup.py develop
```

Thinking more about it, we should probably not bother supporting the case of:

```
    try:
        from skbuild.constants import CMAKE_INSTALL_DIR as SKBUILD_CMAKE_INSTALL_DIR
        _data = os.path.abspath(os.path.join(
            os.path.dirname(__file__), '../../..', SKBUILD_CMAKE_INSTALL_DIR, 'src/python/createrepo_c/data'))
```
and always expect developer to have an editable install. (which is very reasonable requirement)

Then, instead of looking up for binaries in the install tree, a clear message would be reported.

dralley:
Thank you again @jcfr , you've really gone above and beyond :)

I've added one more commit to automatically parse the version string from VERSION.cmake, so that it will not have to be changed in multiple places with every new release.  And I will likely add some instructions to the README later today.

Other than that, @Conan-Kudo, I think it's ready for review.

setup.py
```diff
@@ -1,9 +1,20 @@
 from skbuild import setup
 
+version = '{major}.{minor}.{patch}'
+
+with open('VERSION.cmake', 'r+') as version_file:
+    version_text = version_file.read()
+    # split on lines, remove empty lines
+    lines = filter(len, version_text.split('\n'))
+    # parse out digit characters from the line, convert to int
+    numbers = [int("".join(filter(str.isdigit, line))) for line in lines]
```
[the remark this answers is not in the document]
jcfr: If it would be a pure python package, I would suggest to use [python-versioneer](https://github.com/warner/python-versioneer)

But in the case of `createrepo_c`, the project can be built either as a standalone project or python package, the version management needs to be available to both.

For now, I suggest to stick with your approach. Later after we standardize approaches  like the one done [here](https://github.com/SimpleITK/SimpleITK/blob/master/CMake/sitkSourceVersion.cmake), this could be revisited.

jcfr:
> Thank you again @jcfr , you've really gone above and beyond :)

Glad to see that scikit-build is useful :smile: 

When the createrepo_c package will be ready, I was thinking we could write a blog featuring createrepo_c and scikit-build, would you be up for that ?

dralley:
@jcfr I'm not  a createrepo_c maintainer or anything, just a user who needed a feature, so I'll let the others weigh in on that one.

I wouldn't mind *attempting* to but writing is not really my forte.

jcfr:
> so I'll let the others weigh in on that one.

Sounds good.

> just a user who needed a feature

That is great, i think you should still take part in telling the story. I think it illustrates how community, open-source and transparency lead to great outcome.

README.md
```diff
@@ -102,6 +102,31 @@ cause degradation of performance.
...
+
+To create a binary "wheel" distribution, use:
+
+    python setup.py bdist_wheel
+
+To create a source distribution, use:
+
+    python setup.py sdist
+
+Source distributions require the installer of the package to have all of the build dependencies installed on their system, since they compile the code during installation. Binary distributions are pre-compiled, but they are likely not portable between substantially different systems, e.g. Fedora and Ubuntu.
```
jcfr: `Source distributions require` -> `Installing source distributions require`

README.md
```diff
@@ -102,6 +102,31 @@ cause degradation of performance.
...
+
+If you install a source distribution, you will need
```
jcfr: Look like this sentence is unfinished

README.md
```diff
@@ -102,6 +102,31 @@ cause degradation of performance.
 
 Without git revision specified HEAD is used.
 
+## Build Python package
+
+To create a binary "wheel" distribution, use:
+
+    python setup.py bdist_wheel
```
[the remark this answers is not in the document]
dralley: Are there any specific benefits?  I've not seen this way before, but according to google, it can be fairly slow because it copies everything into a temp directory before doing the build.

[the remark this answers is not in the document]
jcfr: Moving forward, the idea is to build python package using a standard approach that is independent from the build backend. The fact setuptools is used in an implementation details and we should try to move away from it, ideally i think the user experience should be the same for building wheels.

The other advantage is that it keeps the source tree clean of any artifact. 

Timing wise, here is quick comparison for this project

```
$ mkvirtualenv -p python3.5 createrepo_c-pip-wheel

$ time pip wheel . -w dist --no-cache-dir
real	0m5.113s
user	0m13.332s
sys	0m1.104s
```

```
$ mkvirtualenv -p python3.5 createrepo_c-setup-bdist-wheel

$ time bash -c "pip install -r requirements-dev.txt && python setup.py bdist_wheel"
real	0m3.947s
user	0m11.996s
sys	0m1.180s
```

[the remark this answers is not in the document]
dralley: Fair points.  The only objection I have is that w/ Pip v9 (which is what is currently installed on most distributions, including the newest Fedora and Ubuntu), you would also have to install the ```wheel``` package because that functionality isn't included by default.  It looks like Pip v10+ does include it however.

I would probably err on the side of consistency, so if there's an analogous pip command for building sdists, we can use those.  But otherwise I would probably stick with the "standard" way for now and update it later on.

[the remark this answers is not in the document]
jcfr: > so if there's an analogous pip command for building sdists

No

> I would probably stick with the "standard" way for now and update it later on.

Agreed.

dralley:
@jcfr For some reason libdrpm isn't being properly compiled statically, so building on a system where ```drpm-devel``` is present and then installing the resulting wheel on a system where it is not present results in an error when importing ```createrepo_c``` from a Python shell.  It says it cannot find ```libdrpm.so```.

dralley:
Since the Python bindings don't even expose support for deltarpm yet, I'm just going to disable it entirely - problem solved.  See new commit.

Conan-Kudo:
There are a couple of overall issues I have with this:

1. My understanding is that with something like `pyproject.toml` (PEP 518) and buildsys integration (PEP 517), it should be possible for CMake _itself_ to drive the build. This feels completely backwards.

2. This package requires linking to librpm, so how do you deal with that problem?

dralley:
Re: 1

PEP 517 is only "provisionally accepted" and it hasn't been implemented yet.  Until it is (and becomes widespread), everybody still has to use ```setup.py```.  Unfortunately it looks like the implementation is currently stalled out a bit because it's running on volunteer time, so it may take a while before it's ready much less available on enough client systems to switch to.

https://github.com/pypa/pip/issues/5407

FWIW, CMake *is* doing all of the build work.  The only thing that scikit-build does is to implement the existing build hooks, so that when somebody (```pip```, the user) does something with ```setup.py```, it just executes CMake, which performs the build, and then it takes the resulting files and packages them up into a wheel which CMake doesn't know how to do.

Re: 2

Most Python packages that include compiled code are targeted towards an environment called [Manylinux](https://github.com/pypa/manylinux).  It's basically a CentOS 5 image upon which they've built [tooling](https://github.com/pypa/auditwheel) which uses PatchELF to bundle all of the dynamically-linked runtime dependencies into the binary itself, essentially turning all of the binaries statically-linked.

Unfortunately, since it's CentOS 5, the currently existing can't be used to build ```createrepo_c```, since Glib 2.0 is too old and doesn't provide ```gchecksum.h```.  Fortunately, a [new one](https://github.com/pypa/manylinux/pull/182) is in progress based on CentOS 6, and that should work.  Unfortunately, that project is *also* currently running on volunteer time and stalled out.

So for now the best we can do is distribute an sdist, which gets compiled automatically as a bdist at installation time.  This does require the user to have all of the correct build dependencies and libraries installed on their machine (cmake, librpm, etc.)

dralley:
Re: 1, 

Are you referring to how inside ```src/python/CMakeLists.txt``` it is checking whether it's building inside of a scikit-build context and doing slightly different things in that case?  I can see how that might be confusing and look "backwards", but let me explain why it's probably better this way.

Looking at ```src/python/CMakeLists.txt``` [line 1](https://github.com/dralley/createrepo_c/blob/python-pkg/src/python/CMakeLists.txt#L1)

* We can't hardcode ```-DPYTHON_DESIRED=?``` in ```setup.py``` because it has to work for both Python 2 and 3
* If the user is using ```python2 setup.py bdist_wheel```, or is installing it inside of a Python 2 virtual environment, then we automatically know we *must* build against Python 2, and likewise for Python 3.
* It's unnecessary to make the user specify this manually, and possibly invalid if it were specified incorrectly.  Plus, if I'm installing an sdist from PyPI, I don't think those parameters *can* be specified, and even if they can, they shouldn't be.  The user needs to be able to just ```pip install createrepo_c``` and have everything just work.
* Therefore this needs to be context-sensitive, and the best way to do this is just to let ```scikit-build``` figure out what Python libs its supposed to be using (since it knows this information by virtue of being hooked in w/ setuptools, while CMake can't possibly know this information up-front).

To explain the rest of those [instances](https://github.com/dralley/createrepo_c/blob/python-pkg/src/python/CMakeLists.txt#L1)

* The first one is basically the last bullet point mentioned above.  ```PythonExtensions``` is a CMake module provided by ```scikit-build``` that is capable of figuring out what Python libraries to target based on the context of the build.
* The last one is fixing an unsanitary operation -- in the case where we're building a Python package, we *don't* want it to install it to the local environment, we want it to just leave the files somewhere where the scikit-build / setuptools can find them, so that they can package them up into the wheel.
* The middle one was already there prior to this PR, but it's doing something we don't need to do in the Python packaging context, so we skip it.

I hope this explanation makes it sound more reasonable.  
* The second and third ones are basically just saying "install to ```src/python/createrepo_c``` instead of

setup.py
```diff
@@ -0,0 +1,58 @@
...
+        patch=numbers[2]
+    )
+
+setup(
+    name='createrepo_c',
+    description='C implementation of createrepo',
+    version=version,
+    license='GPLv2',
+    author='RPM Software Management',
+    author_email='',
```
Conan-Kudo: Email address is `rpm-ecosystem@lists.rpm.org`

setup.py
```diff
@@ -0,0 +1,58 @@
...
+        major=numbers[0],
+        minor=numbers[1],
+        patch=numbers[2]
+    )
+
+setup(
+    name='createrepo_c',
+    description='C implementation of createrepo',
+    version=version,
+    license='GPLv2',
```
Conan-Kudo: It is `GPLv2+`

src/CMakeLists.txt
```diff
@@ -110,16 +115,34 @@ CONFIGURE_FILE("deltarpms.h.in" "${CMAKE_CURRENT_SOURCE_DIR}/deltarpms.h" @ONLY)
 IF (CMAKE_SIZEOF_VOID_P MATCHES "8")
     SET (LIB_SUFFIX "64")
 ENDIF (CMAKE_SIZEOF_VOID_P MATCHES "8")
+
 SET (LIB_INSTALL_DIR "${CMAKE_INSTALL_PREFIX}/lib${LIB_SUFFIX}")
```
Conan-Kudo: Can we just go ahead and change this to use `GNUInstallDirs`? We're already saying we require at least CMake 2.8, right?

dmach:
I've tested that createrepo_c builds fine with this patch.
I'm merging the PR.
We can solve additional issues as new PRs.

Please note that DNF team has no intention to manage createrepo_c in pypi.

PR #105, opened 2018-10-12
Add option to create repodata reproducibly

marmarek:
Add `createrepo_c --set-timestamp-to-revision`, that set timestamps to the value given with `--revision` option. This way, executing `createrepo_c --revision xxx --set-timestamp-to-revision` twice on the same packages produce bit-by-bit identical results.

This is needed to [reproducibly build](https://reproducible-builds.org/) installation image. See also discussion on similar PR to createrepo, which lead to this specific option: https://github.com/rpm-software-management/createrepo/pull/9

Is there any place where test for `createrepo_c` call as a whole would fit? Since most of the logic is in `main()` (which have `exit()` at the end), it isn't suitable for unit tests...

### DIFF
--- a/src/createrepo_c.c
+++ b/src/createrepo_c.c
@@ -1410,6 +1410,21 @@ deltaerror:
         cr_repomd_record_rename_file(prestodelta_rec, NULL);
     }
 
+    if (cmd_options->set_timestamp_to_revision) {
+        // validated already in cmd_parser.c:check_arguments
+        gint64 revision = strtoll(cmd_options->revision, NULL, 0);
+        cr_repomd_record_set_timestamp(pri_xml_rec, revision);
+        cr_repomd_record_set_timestamp(fil_xml_rec, revision);
+        cr_repomd_record_set_timestamp(oth_xml_rec, revision);
+        cr_repomd_record_set_timestamp(pri_db_rec, revision);
+        cr_repomd_record_set_timestamp(fil_db_rec, revision);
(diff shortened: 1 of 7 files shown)


src/createrepo_c.c
```diff
@@ -1190,6 +1190,21 @@ main(int argc, char **argv)
         cr_repomd_record_rename_file(prestodelta_rec, NULL);
     }
 
+    if (cmd_options->set_timestamp_to_revision) {
+        // validated already in cmd_parser.c:check_arguments
+        gint64 revision = strtoll(cmd_options->revision, NULL, 0);
```
dmach: What about to support --revision and --set-timestamp-to-revision independently?
I think it's a valid use case to specify --set-timestamp-to-revision only when I want to set the revision and timestamps to the current time.

(pseudo-code follows)
```
// use user-defined revision or current time
revision = options->revision ? options->revision : time(NULL))

if options->set_timestamp_to_revision {
   // allways set the timestamps if the option is present (regardless if --revision is specified)
   cr_repomd_record_set_timestamp(...)
   ...
}

```
this may require a small change in cr_xml_dump_repomd_body()
there's a fallback to time(NULL) and it may not be needed any more as repomd->revision will allways be set

marmarek: IMO if you need to control details like revision number, or timestamps, explicit is better than implicit. It's easy enough to specify `--revision $(date +%s)`.

marmarek:
Rebased on master.

PR #108, opened 2018-11-09
Fix setup of logging (log domains) (RhBug:1645897)

jrohel:
New debug messages were added into GLib library. These messages come
from the "GLib" log domain and were not hidden in the standard and
quiet mode of the application.
This fix hides log messages regardless on source log domain.


drizt:
Works for me.

PR #114, opened 2018-12-26
Disable Zchunk for Python package builds

dralley:


Conan-Kudo:
@dralley You're probably going to want zchunk support, since Fedora 30 onward will enable and use it. I know you're using this for Pulp, and it'd be foolish to not be able to create zchunk metadata with Pulp.

Conan-Kudo:
@dralley Do what I do: use a Fedora environment within Travis using the docker image. You can see how I have it set up in rpmlint: https://github.com/rpm-software-management/rpmlint/blob/master/.travis.yml

bmbouter:
+1 to not merging this PR and instead switching Pulp3 RPM's CI to test on Fedora with Travis instead.

PR #115, opened 2019-01-06
Add zchunk support to modifyrepo_c as well as for mergerepo_c and createrepo_c for extra metadata

jdieter:
This patch fixes cr_close() so it doesn't error if zck_end_chunk returns a positive number.


ignatenkobrain:
Spaces after ifs please :)

jdieter:
Done!  :)

jdieter:
@dmach, is there anything else you'd like me to do with this before you review it?

PR #116, opened 2019-01-08
Add module element in updateinfo.xml

kontura:
Two features:
 - add CMake support for python interpreter path, so its possible to either specify python version or path to interpreter
 - add support for module element in updateinfo.xml

### DIFF
--- a/src/updateinfo.h
+++ b/src/updateinfo.h
@@ -51,8 +51,19 @@ typedef struct {
 } cr_UpdateCollectionPackage;
 
 typedef struct {
+    gchar *name;
+    gchar *stream;
+    guint64 version;
+    gchar *context;
+    gchar *arch;
+
+    GStringChunk *chunk;
+} cr_UpdateCollectionModule;
+
+typedef struct {
     gchar *shortname;   /*!< e.g. rhn-tools-rhel-x86_64-server-6.5.aus */
     gchar *name;        /*!< e.g. RHN Tools for RHEL AUS (v. 6.5 for 64-bit x86_64) */
+    cr_UpdateCollectionModule *module;
     GSList *packages;   /*!< List of cr_UpdateCollectionPackage */
     GStringChunk *chunk;
 } cr_UpdateCollection;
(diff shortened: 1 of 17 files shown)


Conan-Kudo:
I'm a little concerned we're just randomly extending the updateinfo.xml format here, especially since createrepo_c cannot validate or create module XML metadata.

If we want to merge this, I _really_ want this guarded by a flag, because it technically generates invalid updateinfo.xml per the schema for it.

dmach:
@Conan-Kudo could you be more specific about the module XML metadata?
What do you mean exactly?
The goal here is to extend updateinfo with module identifiers (N:S:V:C:A)
so we can track relation between streams and RPMs.

About the updateinfo.xml schema - do you know who owns it and where it lives?
We definitely want to stay in sync with the schema.
The updateinfo.xml from F29 doesn't contain any schema link and that makes me think that the format is more a tribal knowledge than a proper standard.

Conan-Kudo:
@dmach I mean that `createrepo_c` can't validate links between the XML data in updateinfo about modules and the modules YAML data to make sure it is valid.

As for the `updateinfo.xml` schema... I know of two schema definitions that exist. One is in [python-Updateinfo](https://cdcvs.fnal.gov/redmine/projects/python-updateinfo), which comes from Scientific Linux. This project attempts to rationalize the variation between Bodhi and OBS's implementations. The other is in [libzypp](https://github.com/openSUSE/libzypp/tree/master/zypp/parser/yum/schema), which I think was used to do the original implementation of the YUM/RPM-MD format for Zypper a decade ago...

dmach:
@Conan-Kudo I think validating links between updateinfo.xml and modulemd is out of scope.
This has to be tested outside createrepo_c.
If you think about it, updateinfo RPMs are not validated against primary.xml either.

Thanks for the schema links.

Conan-Kudo:
@dmach Hmm.

src/updateinfo.h
```diff
@@ -50,9 +50,20 @@ typedef struct {
     GStringChunk *chunk;
 } cr_UpdateCollectionPackage;
 
+typedef struct {
+    gchar *name;
+    gchar *stream;
+    gchar *version;
```
goosemania: @kontura,  is there a reason for a `version` to be of `gchar` type? 
In [libmodulemd](https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/v2/modulemd-module-stream.c#L31) it's `guint64`.

kontura: I have updated it to be `guint64` as well. 
One thing I am not sure about, what should happen when the parsing of the `version` number fails? Is setting it to 0 ok?

PR #117, opened 2019-01-31
Add CMake support for python interpreter path

kontura:
As requested separate pull request for python interpreter path in CMake


dmach:
I've tested a local build with this patch and it passed. Merging.

PR #124, opened 2019-02-11
When authority (host) is present, cr_get_local_metadata would not get path starting with /.

adelton:
RFC 8089 Appendix B shows that the local file with empty authority has three slashes.

That same example appendix also shows `file:/path/to/file` but the original
code did not support it anyway.


adelton:
Fixed typo in the commit message -> 6743495986884c369323195a3366759bd139670a.

PR #126, opened 2019-02-12
mergerepo_c: Consistently produce valid URLs by prepending protocol. (RhBug:1632121)

kontura:
mergerepo_c: Unify url creation when adding package

### DIFF
--- a/src/mergerepo_c.c
+++ b/src/mergerepo_c.c
@@ -537,17 +537,6 @@ cr_srpm_val_destroy(gpointer data)
 }
 
 
-/** Prepend protocol if necessary
- */
-static gchar *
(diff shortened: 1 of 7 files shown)


adelton:
Wouldn't it be easier to just add the protocol (if necessary) when serializing out the file, rather on multiple places? Also, are there any tests in the codebase which would test the behaviour?

kontura:
Yes, I agree that it would be better to add the protocol (if necessary) when serializing, because it would also prevent additional problems with other tools. For example createrepo_c with --update-md-path can now produce paths without protocol.

I will change it and create tests.

dmach:
It compiles and tests pass. Merging.

PR #135, opened 2019-03-08
modifyrepo_c: Prevent doubling of compression (test.gz.gz) (RhBug:1639287)

kontura:
Without this if we try to modifyrepo_c compressed file it ends up compressed twice.

First commit makes modifyrepo_c recognize compression of input file, instead of just assuming its not compressed.
Second commit handles suffix in name of resulting file.


Conan-Kudo:
@kontura Can you rebase this onto current master?

Conan-Kudo:
@kontura Can you please rebase this?

kontura:
In addition to rebasing I have fixed some tests compiler warnings, removed some unnecessary includes and split the unit tests into separate commits.

PR #136, opened 2019-03-12
fix test on armhfp, which does not support printing 64-bit integers

kontura:


pkratoch:
The scratch-build for rawhide passed on all architectures, so I am merging this. Thanks for the work!

Issue #137, opened 2019-03-12
mergerepos_c doesn't support --koji `simple` mode


tkopecek:
We've introduced [new behaviour](https://pagure.io/koji/pull-request/1066) in koji's variant of mergerepo some time ago. This was needed to support some distributions which doesn't have "correct" repos (mixing rpms from different versions of origin srpm). It would be nice to have compatible implementation in mergerepo_c, so users can once again use faster implementation.

Conan-Kudo:
For reference, the new mergerepos changes:
* https://pagure.io/koji/c/a2c61d39924e6768474916837ee91de66b889022
* https://pagure.io/koji/c/9effd55c397b4a1abda33cf0a654ec01a0011b99

kontura:
I created a PR #139 for this, but I am not sure if it covers all the desired behavior.

Conan-Kudo:
This is now supported in v0.13.0.

PR #139, opened 2019-03-19
Add support for --koji simple mode

kontura:
I have tested this on http://download.opensuse.org/distribution/openSUSE-stable/repo/oss/ against koji's mergerepos and there is a difference in inclusion of several packages. The problem are packages which have identical pkgId (hash of the package) as some other package. 

For example there are two identical package files named:
noarch/yast2-trans-de-84.87.20180514.157a0650d-lp150.1.1.noarch.rpm
boot/x86_64/yast2-trans-de.rpm

Basically for this specific repository all the packages in boot/ directory have duplicates.

Is this something we also need to support for issue #137?
I am asking because this is not supported even in the regular --koji mode as they share that part of the code.


tkopecek:
Let's not support it for now. Keep 1:1 with koji's mergerepo behaviour.

tkopecek:
Sorry, that I've not got to it earlier. I've just confirmed, that behaviour is same with both koji/createrepo_c's mergerepos. Error with same hashes, thus ignoring packages is raised in both (in underlying library), so I'm giving +1 here.

kontura:
I have converted long `if () {} else if () {} else if ..` into `switch` statement when selecting merging method, I have also renamed enum values for merge methods to hopefully more reasonable names. 
Newly created header file mergerepo_c.h now contains only necessary definitions and includes.

In addition I have added a new commit, which prevents mergerepo_c from running with undefined command line options.

src/koji.c
```diff
@@ -0,0 +1,284 @@
...
+        }
+
+        if (koji_stuff->blocked_srpms){
+            gboolean is_blocked;
+            is_blocked = g_hash_table_lookup_extended(koji_stuff->blocked_srpms, nevra->name, NULL, NULL);
+            if (is_blocked) {
+                // Srpm of the package is not allowed
+                g_debug("Package %s has blocked srpm %s", pkg->name,
+                        pkg->rpm_sourcerpm);
+                return 0;
```
jrohel: cr_nevra_free(nevra);

src/mergerepo_c.c
```diff
@@ -902,10 +544,10 @@ add_package(cr_Package *pkg,
                                 pkg->name, pkg->arch);
                         return 0;
                     }
+                    break;
 
                 // NVR merge method
-                } else if (merge_method == MM_NVR) {
-
+                case MM_NEWEST_FROM_IDENTICAL_NEVRA: {
```
jrohel: Not identical nevras. Eg. MM_HIGHEST_FROM_NEVRA

kontura:
- Renamed misleading `MM_NEWEST_FROM_IDENTICAL_NEVRA` to `MM_WITH_HIGHEST_NEVRA`
- added `cr_nevra_free(nevra)` calls to prevent memory leaks in `koji_allowed`
- removed some trailing spaces
- removed unused variable `blocked_srpms`

Conan-Kudo:
@kontura Can you rebase onto current master?

kontura:
- added Copyright notices to new files
- moved some more `#includes`  between source and header files

src/mergerepo_c.c
```diff
@@ -220,7 +114,10 @@ static GOptionEntry cmd_entries[] =
 
     // -- Options related to Koji-mergerepos behaviour
     { "koji", 'k', 0, G_OPTION_ARG_NONE, &(_cmd_options.koji),
-       "Enable koji mergerepos behaviour.", NULL},
+       "Enable koji mergerepos behaviour. (Optionally select simple mode with: --simple)", NULL},
+    { "simple", 0, 0, G_OPTION_ARG_NONE, &(_cmd_options.koji_simple),
+       "Enable koji specific simple merge mode where we keep even packages with "
+       "identical NEVRAs. (Doesn't work without --koji)", NULL},
```
jrohel: There is new text `(Doesn't work without --koji)` added.
There is another option `blocked` with text `Only works with combination with --koji/-k.`
For consistency I prefer the same message for both options.

kontura: Pushed an update with reformatted option text.

PR #142, opened 2019-04-01
Make createrepo obsolete conditional

j-mracek:
The obsolete should be applied only for Fedora 31+ where createrepo is
going to disappear, and for RHEL8.

The obsolete in EPEL-7 cannot be applied, because it obsoletes the core
package of RHEL7.

### DIFF
--- a/createrepo_c.spec
+++ b/createrepo_c.spec
@@ -61,8 +61,11 @@ Requires: rpm >= 4.9.0
 %if %{with drpm}
 BuildRequires:  drpm-devel >= 0.1.3
 %endif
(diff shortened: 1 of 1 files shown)


bleve:
I don't think that is enough. There is still /usr/bin/* symlinks which conflict with createrepo - I think those should not be intalled on older systems.

pkratoch:
Thanks for pointing that out. I made a PR that adds conflict with createrepo if the obsolete is not there: https://github.com/rpm-software-management/createrepo_c/pull/144

PR #144, opened 2019-04-01
Make createrepo, mergerepo, modifyrepo symlinks conditional

pkratoch:


createrepo_c.spec
```diff
@@ -65,6 +65,8 @@ BuildRequires:  drpm-devel >= 0.1.3
 %if 0%{?fedora} > 30 || 0%{?rhel} > 7
 Obsoletes:      createrepo < 0.11.0
 Provides:       createrepo = %{version}-%{release}
+%else
+Conflicts:      createrepo
```
j-mracek: What about "createrepo < 0.11.0"

j-mracek: But I am not sure.

Issue #131, opened 2019-03-05
mergerepo needs to support merging module metadata


sgallagher:
We need to handle cases where a user has built a module repository (such as by using MBS local builds) and wants to add it to an existing repo (such as a company-internal repository).

What mergerepo_c needs to do is this:
* Link against libmodulemd
* Use libmodulemd routines to read in the modules.yaml from the source repo
* Read in the modules.yaml from the additive repo(s)
* Use the ModuleIndexMerger functionality to merge the two sets of metadata
* Write the metadata YAML back out and store it.

Documentation: https://sgallagh.fedorapeople.org/docs/libmodulemd/latest/modulemd-2.0-Modulemd.ModuleIndexMerger.html

Conan-Kudo:
This should be supported now in v0.13.0.

Issue #134, opened 2019-03-06
Add native support for module metadata


sgallagher:
Rather than using the `--mdtype` argument to `modifyrepo_c`, it needs to be possible to have createrepo_c accept a modulemd as standard input. This will be a necessary prerequisite to #132

sgallagher:
Quite a bit of discussion on this has occurred during the code review of #146 with @Conan-Kudo 

Capturing some of it here for future reference:

> @Conan-Kudo Would you mind spelling out what you see as the validation cases (and what should happen if there's a discrepancy)?
> 
> Here are the cases I can think of:
> 
> ## Prerequisites
> A modulemd YAML document is provided to createrepo_c that contains a mixture of zero or more
> * Module Stream documents
> * Module Default documents
> * Module Translation documents
> 
> ## Case 1: Entirely valid modulemd
> Conditions:
> * Module Stream documents: all entries in the `data.artifacts.rpms` section match RPM filenames in the repository being manipulated
> * Module Default documents: all defaults reference a module for which there is at least one Module Stream document in the stream.
> * Module Translation documents: all translations reference a module for which there is at least one Module Stream document in the stream.
> 
> This is the absolute easiest case. This modulemd YAML should be preserved in its entirety.
> 
> ## Case 2: Some missing artifacts
> Conditions:
> * Module Stream documents: One or more of these documents have some entries in the `data.artifacts.rpms` section that are not reflected in the repository and some that are.
> * Module Default documents: all defaults reference a module for which there is at least one Module Stream document in the stream.
> * Module Translation documents: all translations reference a module for which there is at least one Module Stream document in the stream.
> 
> Do we filter out this module stream object or assume that as long as at least one item in the artifacts list is present that we should keep this stream document? I think the latter, since to do otherwise means that such artifacts would then be viewed by DNF as if they were non-modular and therefore considered for updates where they should not be.
> 
> ## Case 3: All missing artifacts
> Conditions:
> * Module Stream documents: One or more of these documents have their `data.artifacts.rpms` section containing no artifacts that match the contents of the repository.
> * Module Default documents: all defaults reference a module for which there is at least one Module Stream document in the stream.
> * Module Translation documents: all translations reference a module for which there is at least one Module Stream document in the stream.
> 
> In this case, do we filter the offending Module Stream document(s) out of the final output?
> 
> ## Case 4: Defaults with no matching stream
[...]

PR #146, opened 2019-04-03
Add support for reading and merging module metadata

sgallagher:
This is the first step towards addressing https://github.com/rpm-software-management/createrepo_c/issues/131

Since this is my first time contributing to this project, I'm going to be publishing my work-in-progress to this PR as I go, looking for feedback as to whether my approach makes sense.


CMakeLists.txt
```diff
@@ -39,6 +39,7 @@ find_package(EXPAT REQUIRED)
 find_package(GLIB2 REQUIRED)
 find_package(GTHREAD2 REQUIRED)
 find_package(Libmagic REQUIRED)
+find_package(LIBMODULEMD REQUIRED)
```
dmach: My only concern here is that libmodulemd is a required build dependency.
Some distros don't care about modularity and we should allow them to build without libmodulemd.

I recommend adding following option:
OPTION(WITH_MODULEMD "Enables modulemd support" ON)

sgallagher: I discussed this with @dmach over IRC and we came to the conclusion that libmodulemd will be made optional, but that createrepo_c will refuse to operate on repos containing module metadata if it was not compiled with libmodulemd support. This is the only safe course of action, since inadvertently dropping module metadata can cause a great number of issues.

sgallagher:
I've pushed a fixup patch (for ease of review comparison) that makes libmodulemd optional and adds code (and a test) to have `cr_locate_metadata()` return an error if it discovers a repository with module metadata and libmodulemd support was not compiled in.

I also switched the tests to use a new REPO_03, rather than modifying REPO_00 with module metadata because the existing tests would now have failed if libmodulemd wasn't compiled in.

I'll squash the patches together once they are approved.

@dmach @kontura

dralley:
The README should probably be updated to note the new dependency.

sgallagher:
@dralley Thanks, good catch. I just pushed another fixup patch to address that.

tests/test_load_metadata.c
```diff
@@ -151,6 +174,31 @@ static void test_cr_metadata_locate_and_load_xml_detailed(void)
...
+    guint size;
+    cr_Package *pkg;
+    cr_Metadata *metadata;
+
+    metadata = cr_metadata_new(CR_HT_KEY_NAME, 0, NULL);
+    g_assert(metadata);
+    ret = cr_metadata_locate_and_load_xml(metadata, TEST_REPO_03, NULL);
+    g_assert_cmpint(ret, ==, CRE_OK);
+    size = g_hash_table_size(cr_metadata_hashtable(metadata));
+    g_assert_cmpuint(size, ==, REPO_SIZE_00);
```
kontura: Should be REPO_SIZE_03, but other than this small detail it all looks good to me.

sgallagher:
I fixed the `REPO03` mistake and squashed those patches together. I also just pushed another fixup patch because while looking at modifyrepo_c, I realized that `cr_metadata_load_xml()` is used directly in many places, so I needed to move the module loading into that function instead, or risk losing the module metadata.

The other patch I just pushed updates the `--update` logic to exclude the module metadata from the blacklist so that it doesn't get removed there. I didn't modify the `cr_repodata_blacklist_classic()` variant because it doesn't look like it will catch the modulemd.yaml in its net. I'm not sure what to do about the `cr_repodata_blacklist_by_age()` function, though. I'd prefer to exclude modules.yaml from it as well, but I don't have anything at my disposal except for the suffix of the path, which theoretically could be something other than `modules.yaml`.

sgallagher:
OK, I just pushed a new set of patches to include support for mergerepo_c to merge repositories with (and without) module metadata.

I'd appreciate a review. I *think* this is close to being ready to merge, but please don't merge it right away. I had to write some new features for libmodulemd and want to release that just before this lands. (But I'm waiting until I know for sure there aren't other new features needed for this.)

cmake/Modules/FindLIBMODULEMD.cmake
```diff
@@ -0,0 +1,40 @@
...
+
+	PKG_SEARCH_MODULE ( LIBMODULEMD ${_pkgconfig_REQUIRED} modulemd-2.0>=2.3 )
+
+	IF (LIBMODULEMD_FOUND)
+    IF (NOT LIBMODULEMD_FIND_QUIETLY)
+      MESSAGE (STATUS "Found libmodulemd: ${LIBMODULEMD_LIBRARIES} ${LIBMODULEMD_INCLUDE_DIRS}")
+    ENDIF (NOT LIBMODULEMD_FIND_QUIETLY)
+  ELSE (LIBMODULEMD_FOUND)
+    IF (LIBMODULEMD_FIND_REQUIRED)
+      MESSAGE (SEND_ERROR "Could not find GLib2")
```
ignatenkobrain: GLib2?

cmake/Modules/FindLIBMODULEMD.cmake
```diff
@@ -0,0 +1,40 @@
+# Try to find libmodulemd 2.3+
```
ignatenkobrain: is there any reason to not just use PKG_SEARCH_MODULE directly in CMakeLists.txt?

cmake/Modules/FindLIBMODULEMD.cmake
sgallagher: Copy-paste error. I’ll fix it. Thanks

cmake/Modules/FindLIBMODULEMD.cmake
sgallagher: Honestly, I knew nothing about CMake prior to writing this patch. I just copied the GLib2 approach and trimmed it down to only what we needed from libmodulemd. If that's a sufficient answer (and preferred), I can whittle it down further.

cmake/Modules/FindLIBMODULEMD.cmake
sgallagher: Pushed a fix for this.

CMakeLists.txt
sgallagher: This was implemented in the latest patches. Please re-review.

README.md
```diff
@@ -102,6 +103,12 @@ cause degradation of performance.
 
 Build with zchunk support (Default: ON)
 
+### ``-DWITH_LIBMODULEMD=ON``
+
+Build with libmodulemd support (Default: ON)
+
+Adds support for working with repos containing Fedora Modularity metadata.
```
Conan-Kudo: Could you please provide a link here to Fedora Modularity stuff? No one is going to know what that means otherwise...

src/metadata_internal.h
```diff
@@ -0,0 +1,43 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301,
+ * USA.
+ */
+
+#ifndef __C_CREATEREPOLIB_METADATA_INTERNAL_H__
+#define __C_CREATEREPOLIB_METADATA_INTERNAL_H__
```
Conan-Kudo: Is this file supposed to have more than just modulemd stuff? Because if not, this is probably the wrong name for the file...

tests/testdata/repo_03/repodata/repomd.xml
```diff
@@ -0,0 +1,66 @@
...
+    <checksum type="sha256">a939c4765106655c3f7a13fb41d0f239824efa66bcd6c1e6c044a854012bda75</checksum>
+    <open-checksum type="sha256">e0ed5e0054194df036cf09c1a911e15bf2a4e7f26f2a788b6f47d53e80717ccc</open-checksum>
+    <header-checksum type="sha256">482e0dcc2e58d0e616f9b00c7789948e6c41f89488e7f4ad2fc3bcc7b549c6d5</header-checksum>
+    <location href="repodata/a939c4765106655c3f7a13fb41d0f239824efa66bcd6c1e6c044a854012bda75-other.xml.zck"/>
+    <timestamp>1533242352</timestamp>
+    <size>263</size>
+    <open-size>121</open-size>
+    <header-size>131</header-size>
+  </data>
+  <data type="modules">
```
Conan-Kudo: I'm pretty sure modules can be `zck` compressed too, so we'll want tests for that.

src/error.h
```diff
@@ -96,6 +96,8 @@ typedef enum {
         (33) Cannot change blocked signals */
     CRE_ZCK, /*!<
         (34) ZCK library related error */
+    CRE_MMD, /*!<
```
Conan-Kudo: `CRE_MMD` -> `CRE_MODULEMD`.

Explanatory names are better.

src/helpers.c
```diff
@@ -246,8 +246,15 @@ cr_repodata_blacklist(const char *repodata_path,
...
-                                     g_path_get_basename(rec->location_href));
+        if (g_strcmp0(rec->type, "modules") != 0) {
+            // Never blacklist the modules YAML
+            // This is a stopgap measure until createrepo_c learns to generate
+            // the modulemd.yaml itself. In the meantime, we just need to
+            // ensure that it never deletes the file.
+            *blacklist =
+                g_slist_prepend(*blacklist,
+                                g_path_get_basename(rec->location_href));
+        }
```
Conan-Kudo: I don't think this is a good idea. There are valid use-cases for being able to decompose a modular repo into regular rpm-md ones, or constructing a custom subset of modules in a repo.

sgallagher: Can you be more specific about what you'd rather see? Right now, passing `--update` results in modules.yaml being deleted completely. This at least ensures that they aren't lost.

I'm not clear on what it should be doing instead.

Conan-Kudo: I think the point of this PR is to give `createrepo_c` awareness of what to do with `modules.yaml`. That means I would expect no logic like this to exist in the final form in _absolute_ and _unconditional_ terms. That said, if a switch was added for preserving `modules.yaml` (like we have for other auxiliary metadata), then I'd be okay with it.

sgallagher:
Before I make any more changes to this, can at least two of @dmach @Conan-Kudo and @kontura please confirm that they agree with the requirements I listed in https://github.com/rpm-software-management/createrepo_c/pull/146#discussion_r273532554 as my operating principles?

src/helpers.c
Conan-Kudo: Unfortunately, I have to disagree, because one of the fundamental problems is that the file needs to be validated for components existing in the repo. If they don't exist, it needs to do something about it. It can't treat it as immutable or always correct either, because it's not the source of truth of the repository, createrepo_c is.

sgallagher: @Conan-Kudo Would you mind spelling out what you see as the validation cases (and what should happen if there's a discrepancy)?

Here are the cases I can think of:

## Prerequisites
A modulemd YAML document is provided to createrepo_c that contains a mixture of zero or more
* Module Stream documents
* Module Default documents
* Module Translation documents

## Case 1: Entirely valid modulemd
Conditions:
* Module Stream documents: all entries in the `data.artifacts.rpms` section match RPM filenames in the repository being manipulated
* Module Default documents: all defaults reference a module for which there is at least one Module Stream document in the stream.
* Module Translation documents: all translations reference a module for which there is at least one Module Stream document in the stream.

This is the absolute easiest case. This modulemd YAML should be preserved in its entirety.

## Case 2: Some missing artifacts
Conditions:
* Module Stream documents: One or more of these documents have some entries in the `data.artifacts.rpms` section that are not reflected in the repository and some that are.
* Module Default documents: all defaults reference a module for which there is at least one Module Stream document in the stream.
* Module Translation documents: all translations reference a module for which there is at least one Module Stream document in the stream.

Do we filter out this module stream object or assume that as long as at least one item in the artifacts list is present that we should keep this stream document? I think the latter, since to do otherwise means that such artifacts would then be viewed by DNF as if they were non-modular and therefore considered for updates where they should not be.

## Case 3: All missing artifacts
Conditions:
* Module Stream documents: One or more of these documents have their `data.artifacts.rpms` section containing no artifacts that match the contents of the repository.
* Module Default documents: all defaults reference a module for which there is at least one Module Stream document in the stream.
* Module Translation documents: all translations reference a module for which there is at least one Module Stream document in the stream.

In this case, do we filter the offending Module Stream document(s) out of the final output?

## Case 4: Defaults with no matching stream
* Module Stream documents: all entries in the `data.artifacts.rpms` section match RPM filenames in the repository being manipulated
* Module Default documents: One or more such documents do not reference a module stream present in the metadata.
* Module Translation documents: all translations reference a module for which there is at least one Module Stream document in the stream.

In this situation, I think we still need to preserve the defaults, because this may be a metadata-only change (such as a case where the `fedora` repository contains a module stream and later we modify the stream defaults to treat that stream as the default. This doesn't require a rebuild of the module stream, so it may not be present in the updates repo.

## Case 5: Translations with no matching stream
* Module Stream documents: all entries in the `data.artifacts.rpms` section match RPM filenames in the repository being manipulated
* Module Default documents: all defaults reference a module for which there is at least one Module Stream document in the stream.
* Module Translation documents: one or more translations reference modules not represented in this YAML stream.

In this situation, I think we still need to preserve the translations, because this may be a metadata-only change (such as a case where the `fedora` repository contains a module stream and the only change to it is to add new translations. This doesn't require a rebuild of the module stream, so it may not be present in the updates repo.

sgallagher:
I just pushed a rebased set of patches, dropping the patch that retained the metadata on `--update` for now. Can I get a full review and hopefully a merge of the mergerepo_c changes? We need them as a dependency for Koji changes to support modular packages in the non-modular buildroot.

tests/testdata/repo_03/repodata/repomd.xml
sgallagher: Do we need the zck content in the initial patch, or do I need to figure out how that works and get that done in this patch-set?

src/metadata_internal.h
sgallagher: I named it based on the conventions for other files. It can theoretically have any private internal function definitions needed by createrepolib_metadata. Currently we only need it for module stuff.

sgallagher:
I confirmed with @j-mracek that DNF should handle zchunked modulemd already, so I pushed a patch to add that functionality as well.

I changed `CRE_MMD` -> `CRE_MODULEMD` as requested by @Conan-Kudo 

I didn't bother changing the CMake stuff mentioned by @ignatenkobrain as while it might be overkill, it's at least there in case we need to adjust it later.

I'd appreciate a re-review as soon as possible, please.

src/helpers.c
Conan-Kudo: @sgallagher

Case 1 is fine. My concern with case 2 is that I don't know how it would be handled for modulemd aware reposyncs. In my view, if a component is missing, then it should be dropped from the metadata, because otherwise it'll look like the component is there when it isn't.

Cases 3, 4, and 5 should result in the module document getting dropped. What is the point of keeping a document that has no underlying content? It bloats the metadata for essentially no reason.

sgallagher: > @sgallagher
> 
> Case 1 is fine. My concern with case 2 is that I don't know how it would be handled for modulemd aware reposyncs. In my view, if a component is missing, then it should be dropped from the metadata, because otherwise it'll look like the component is there when it isn't.

The problem with dropping the metadata in that situation is that we may end up with a repo that has modular RPMs that are treated by DNF as non-modular RPMs (meaning that stream filtering doesn't happen). This *will* cause issues.

Maybe we should just agree this is a hard failure for the createrepo operation and that if we end up with only a partial match, we assume that the repo was not properly populated. Reasonable?

> 
> Cases 3, 4, and 5 should result in the module document getting dropped. What is the point of keeping a document that has no underlying content? It bloats the metadata for essentially no reason.

Case 3, I agree with. If we have zero of the artifacts referenced for a stream, let's just drop that from the metadata. That will make life easier for filtering things out if we want to trim a repository.

For cases 4 and 5, I tried to explain in the original comment, but my view is that these are *valid* cases that we cannot reasonably be able to detect, short of duplicating data unnecessarily. It is completely reasonable that we might ship nodejs:10 as a module in Fedora 30's frozen `fedora` repository. Then, later in F30's life, we complete the default-modules-in-the-non-modular-buildroot feature (which this patch is a part of...) and we then push a metadata-only change to the modulemd that marks the nodejs:10 module as the default for Fedora 30, which takes over the update path from the non-modular RPM. If we require that the metadata must have a matching module stream, then it necessitates either rebuilding that module stream for the `fedora-updates` repo or else copying the data from the frozen `fedora` repo into the `fedora-updates` repo, wasting a lot of space. I'd much rather that we just always pass modulemd-defaults documents as-is. I think it's going to be pretty rare that this matters, but if we need to make a change like this and an overzealous filter gets in the way, it'll be difficult to track down and quite annoying.

Case 5 (translations), I think is even *more* likely to hit this than defaults, since new translations could arrive in the compose at any time, whereas defaults changes will be likely less common. We're working on automation to have pungi pull the latest translations from Zanata on each nightly compose and add the metadata to the modulemd document. If createrepo_c then chucks it back out, it will make that function largely useless.

Conan-Kudo: > Maybe we should just agree this is a hard failure for the createrepo operation and that if we end up with only a partial match, we assume that the repo was not properly populated. Reasonable?

That makes sense to me.

> For cases 4 and 5, I tried to explain in the original comment, but my view is that these are valid cases that we cannot reasonably be able to detect, short of duplicating data unnecessarily. It is completely reasonable that we might ship nodejs:10 as a module in Fedora 30's frozen fedora repository. Then, later in F30's life, we complete the default-modules-in-the-non-modular-buildroot feature (which this patch is a part of...) and we then push a metadata-only change to the modulemd that marks the nodejs:10 module as the default for Fedora 30, which takes over the update path from the non-modular RPM. If we require that the metadata must have a matching module stream, then it necessitates either rebuilding that module stream for the fedora-updates repo or else copying the data from the frozen fedora repo into the fedora-updates repo, wasting a lot of space. I'd much rather that we just always pass modulemd-defaults documents as-is. I think it's going to be pretty rare that this matters, but if we need to make a change like this and an overzealous filter gets in the way, it'll be difficult to track down and quite annoying.

Ugh, I see your point...

> Case 5 (translations), I think is even more likely to hit this than defaults, since new translations could arrive in the compose at any time, whereas defaults changes will be likely less common. We're working on automation to have pungi pull the latest translations from Zanata on each nightly compose and add the metadata to the modulemd document. If createrepo_c then chucks it back out, it will make that function largely useless.

I don't think translations for things that don't have any content associated with them would matter that much...

sgallagher: @Conan-Kudo As this is now going into design of another feature, I'm moving this conversation to #134 

Please continue over there.

sgallagher:
Thanks for the review and approvals. Can we get this merged and released fairly soon to unblock the Koji work that depends on it?

sgallagher:
I released libmodulemd 2.3.0 with the new interfaces to support this last night and updates are on their way to Fedora and COPR for EPEL 7.

Can we please get this PR merged?

PR #147, opened 2019-04-10
Option --keep-all-metadata is not limited to groupfile and updateinfo (RhBug: 1670651)

kontura:
With this PR options `--update --keep-all-metadata` preserve all metadata from repomd.xml except those whose names start with: "primary"/"filelists"/"other". 
This however means that when we run `--update --keep-all-metadata` on repository with zck compression we keep the zck compressed files (such as updateinfo.xml.zck) even if we didn't specify the --zck option.


kontura:
Updated generation of .zck files so that we only create one .zck version and not something like updateinfo.xml.gz.zck

Conan-Kudo:
@kontura Can you rebase this?

dmach:
@kontura good point about not keeping the primary_zck etc.
We need to skip zck and also db (sqlite) metadata.

I'd probably skip following:
- primary
- primary_*
- filelists
- filelists_*
- others
- others_*

kontura:
- Rebased
- Removed commit with acceptance tests and converted them to tests for ci-dnf-stack
- Added three new commits mostly fixing problems I ran into while rebasing, because I had to do some changes due to added support for libmodulemd.

doc/createrepo_c.8
```diff
@@ -151,7 +151,7 @@ Which compression type to use.
 Which compression type to use (even for primary, filelists and other xml).
 .SS \-\-keep\-all\-metadata
 .sp
-Keep groupfile and updateinfo from source repo during update.
+Keep all additional metadata (not primary, filelists and other xmls, neither theirs compressed variants) from source repo during update.
```
Conan-Kudo: `neither theirs` -> `nor their`

Also, don't forget to fix the information in the code, since `createrepo_c --help` has to agree too.

kontura: Updated, thank you. :+1:

PR #148, opened 2019-04-12
Don't try to parse directories with ".rpm" suffix as packages during directory walk (RhBug:1348215)

kontura:


acceptance_tests/tests/base.py
```diff
@@ -367,3 +368,22 @@ def assert_same_dir_content(self, a, b):
...
+
+        if path_to_file.endswith(".gz"):
+            with gzip.open(path_to_file, 'rb') as f:
+                for l in f.readlines():
+                    if (l.strip() == line.strip()):
+                        found = True
+        self.assertTrue(found, "Line: " + line + " not found in file: " + path_to_file)
+        return
+
+        with open(path_to_file) as f:
```
jrohel: Code after return? Dead code?

kontura: Oh, my mistake, I have updated it.
Thank you.

PR #151, opened 2019-04-23
Prevent race condition when --update & we have duplicate pkgs (RhBug:1696808)

kontura:
This PR aims to fix possible race condition when we use `--update` option (which loads old metadata but identical pkgs get only one record) and our new repo contains multiple occurrences of one pkg. 
Each occurrence of the one pkg is handled in separate thread and they all want to modify `location_href` and `location_base` of the old loaded single record.

With this change only the first thread gets the old record and every other duplicate thread has to load the package from disk again. This approach should have negligible slow down effect on runs with only a few or none duplicate pkgs, which I assume are much more common.


Conan-Kudo:
@kontura Typo: `duplicit` -> `duplicate`

PR #152, opened 2019-04-24
Add libmodulemd dependency to spec file

kontura:
There is a condition to build with libmodulemd only on fedora >= 30 because on 29 libmodulemd 2.3.0 is available only in testing repo.

### DIFF
--- a/createrepo_c.spec
+++ b/createrepo_c.spec
@@ -25,6 +25,12 @@
 %bcond_without zchunk
 %endif
 
+%if 0%{?rhel} || 0%{?fedora} < 29
+%bcond_with libmodulemd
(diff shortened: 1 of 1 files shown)


createrepo_c.spec
```diff
@@ -25,6 +25,12 @@
 %bcond_without zchunk
 %endif
 
+%if 0%{?rhel} || 0%{?fedora} < 30
```
Conan-Kudo: It's going to make its way into stable now: https://bodhi.fedoraproject.org/updates/FEDORA-2019-05908109f7

kontura:
Updated the condition

PR #154, opened 2019-05-03
Add --pkgorigins mode

sgallagher:
When handling modules, we want a mode of operation for Koji that
handles the pkgorigins file but does not do any of the other
filtering or blocking that the --koji mode does.

Note: this patch doesn't need to change any code where the
pkgorigins file is actually populated, because it was already
conditionalized merely on the koji_stuff->pkgorigins pointer
being non-NULL, not driven by the cmd_options->koji.

Also includes a small patch to make the `koji_stuff_destroy()`
routine more resilient and handle when it's not fully constructed.

### DIFF
--- a/doc/mergerepo_c.8
+++ b/doc/mergerepo_c.8
@@ -1,6 +1,6 @@
 .\" Man page generated from reStructuredText.
 .
-.TH MERGEREPO_C  "2017-02-23" "" ""
+.TH MERGEREPO_C  "2019-05-03" "" ""
 .SH NAME
 mergerepo_c \- Merge multiple rpm-md format repositories together
 .
@@ -81,6 +81,9 @@ Don\(aqt add a baseurl to packages that don\(aqt have one before.
 .SS \-k \-\-koji
 .sp
 Enable koji mergerepos behaviour.
+.SS \-\-pkgorigins
+.sp
+Enable standard mergerepos behavior while also providing the pkgorigins file for koji.
 .SS \-g \-\-groupfile GROUPFILE
 .sp
 Path to groupfile to include in metadata.
--- a/src/mergerepo_c.c
+++ b/src/mergerepo_c.c
@@ -118,6 +118,10 @@ static GOptionEntry cmd_entries[] =
     { "simple", 0, 0, G_OPTION_ARG_NONE, &(_cmd_options.koji_simple),
        "Enable koji specific simple merge mode where we keep even packages with "
        "identical NEVRAs. Only works with combination with --koji/-k.", NULL},
+    { "pkgorigins", 0, 0, G_OPTION_ARG_NONE, &(_cmd_options.pkgorigins),
+        "Enable standard mergerepos behavior while also providing the "
+        "pkgorigins file for koji.",
+        NULL},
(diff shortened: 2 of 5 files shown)


sgallagher:
Note: this differs from `--koji --simple` in that it doesn't execute any of the koji-style merge logic. It was considerably harder to conditionalize that, so I opted for it to be its own separate flag. If `--koji` is passed, it ends up being a proper superset of this functionality, so that simply takes precedence.

sgallagher:
Since I realize I didn't justify this originally. This is paired with https://pagure.io/koji/pull-request/1411 for Koji and will be used to generate buildroots containing modular content for non-modular builds.

src/mergerepo_c.c
```diff
@@ -118,6 +118,10 @@ static GOptionEntry cmd_entries[] =
     { "simple", 0, 0, G_OPTION_ARG_NONE, &(_cmd_options.koji_simple),
        "Enable koji specific simple merge mode where we keep even packages with "
        "identical NEVRAs. Only works with combination with --koji/-k.", NULL},
+    { "pkgorigins", 0, 0, G_OPTION_ARG_NONE, &(_cmd_options.pkgorigins),
+        "Enable standard mergerepos behavior while also provsidinf the "
```
Conan-Kudo: `provsidinf` -> `providing`?

Conan-Kudo:
@sgallagher Can you please regenerate the man pages so that the new flags are documented there? The `gen_manpage.sh` script in the `utils` folder should help here...

src/mergerepo_c.c
sgallagher: Typing while exhsuted :-P

sgallagher:
> @sgallagher Can you please regenerate the man pages so that the new flags are documented there? The `gen_manpage.sh` script in the `utils` folder should help here...

Done and pushed. I split that into a separate patch. Note that it changed more than just my flag.

doc/mergerepo_c.8
```diff
@@ -42,9 +42,9 @@ Show program\(aqs version number and exit
 .SS \-r \-\-repo REPOS
 .sp
 Repo url
-.SS \-a \-\-archlist ARCHLIST
+.SS \-\-repo\-prefix\-search ARCHLIST
 .sp
-Defaults to all arches \- otherwise specify arches
+Repository prefix to be replaced by repo\-prefix\-url. }, { repo\-prefix\-replace, 0, 0, G_OPTION_ARG_STRING, &(_cmd_options.repo_prefix_replace), Repository prefix URL by which the repo\-prefix is replaced. }, { archlist, \(aqa\(aq, 0, G_OPTION_ARG_STRING, &(_cmd_options.archlist), Defaults to all arches \- otherwise specify arches
```
Conan-Kudo: Crap, this looks broken... :(

sgallagher: OK, what should I do about it? Should I drop the manpage patch?

Conan-Kudo: For now, I guess. Our only include the bit for the new option.

sgallagher:
Just got back to this. I revised the manpage patch to include only the changes related to this change. Please re-review.

Conan-Kudo:
@sgallagher Changes look good to me, I'll test and merge this shortly.

PR #158, opened 2019-05-09
Fix crash when dumping updateinfo and module is ommited (RhBug:1707981)

kontura:
Also fixed indentation of the whole method and added tests for this case.


Conan-Kudo:
@kontura did you make sure the tests work properly when modulemd support is disabled?

kontura:
I actually didn't I will have to be more careful about that.
I ran them now and it seems fine or do some of them fail for you?

Conan-Kudo:
@kontura I was checking if you had tested it since we had that problem recently. I haven't had a chance to test yet myself...

PR #163, opened 2019-05-29
Depend on the appropriate minimum version of libmodulemd

pkratoch:
Without this, under some circumstances it's not updating libmodulemd on the
system and if it gets 2.2.x or older, createrepo_c fails with undefined
symbol: modulemd_module_index_update_from_custom

Based on commit in Fedora downstream: https://src.fedoraproject.org/rpms/createrepo_c/c/40377017d0fc34d222a56ea1ad07ed63557a7a7d

### DIFF
--- a/createrepo_c.spec
+++ b/createrepo_c.spec
@@ -1,3 +1,5 @@
+%global libmodulemd_version 2.3.0
+
 # Bash completion (we need different approach for RHEL-6)
 %if 0%{?rhel} == 6
 %global bash_completion %config%{_sysconfdir}/bash_completion.d/createrepo_c.bash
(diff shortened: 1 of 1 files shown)


createrepo_c.spec
```diff
@@ -60,6 +60,7 @@ BuildRequires:  zchunk
 %if %{with libmodulemd}
 BuildRequires:  pkgconfig(modulemd-2.0) >= 2.3.0
 BuildRequires:  libmodulemd
+Requires:       libmodulemd%{?_isa} >= 2.3.0
```
Conan-Kudo: We should probably change this to have the version value controlled by a macro, since it needs to be the same at build-time and runtime.

pkratoch: Fixed.

Issue #104, opened 2018-10-12
 createrepo_c: non printable character in provides tag leads to invalid xml


M0ses:
As reported by @lnussel in

https://bugzilla.opensuse.org/show_bug.cgi?id=1110914

An rpm with a non printable character in it's provides tag leads createrepo_c to create invalid XML which makes other tools barf:

<rpm:entry name="ksym(default:^B)"/>

lnussel:
what's the status here?

M0ses:
ping!

kontura:
I have been trying to solve this issue, but I got stuck.
I am not sure how should createrepo_c behave when this happens.

Original createrepo just ignored the characters and simply dropped them, this however leads to wrong repodata and DNF could have trouble with packages like this (it depends on where the character was in the original rpm, but dropping characters from file names, provides, requires... not good)

On the other hand I tried escaping these characters but this still leads to invalid xmls. Basically every tool I tried on such repodata (xmllint/xmlwf) complains. DNF ignores such a repo entirely, just like when the characters are no escaped. 
According to  [this document](https://www.w3.org/TR/xml/#charsets) only #x9, #xA and #xD (tab, line feed and carriage return) are allowed from non printable characters in the text data of xml.

The only option that I can think of is not including packages with such characters in the resulting repodata.
Or am I wrong somewhere? Is there some other way we could do this?
I would be grateful for any help.

M0ses:
> I have been trying to solve this issue, but I got stuck.
> I am not sure how should createrepo_c behave when this happens.
> 
> Original createrepo just ignored the characters and simply dropped them, this however leads to wrong repodata and DNF could have trouble with packages like this (it depends on where the character was in the original rpm, but dropping characters from file names, provides, requires... not good)

I fully agree!

> 
> On the other hand I tried escaping these characters but this still leads to invalid xmls. Basically every tool I tried on such repodata (xmllint/xmlwf) complains. DNF ignores such a repo entirely, just like when the characters are no escaped.
> According to [this document](https://www.w3.org/TR/xml/#charsets) only #x9, #xA and #xD (tab, line feed and carriage return) are allowed from non printable characters in the text data of xml.
> 
> The only option that I can think of is not including packages with such characters in the resulting repodata.
> Or am I wrong somewhere? Is there some other way we could do this?

IMHO createrepo_c should exit with an error as early as possible to make this error visible to "users" aka package maintainers. Otherwise it might become really hard to find the root case.

[...]

PR #165, opened 2019-06-07
Don't include packages with forbidden control chars in repodata + some warning fixes

kontura:
As discussed in https://github.com/rpm-software-management/createrepo_c/issues/104 this PR ensures that packages with forbidden control chars are not included in repodata and appropriate warning is generated. Also when `--error-exit-val` is used and such a package is encountered return value is set to 2 as with other errors.

The third commit addresses compiler warnings when cmake is ran with -DCMAKE_BUILD_TYPE:STRING=DEBUG. Removing various unused code, fixing some incompatible types, initialization and so on.


sergiomb2:
IMO     ` g_set_error(err, CREATEREPO_C_ERROR, CRE_XMLDATA,
                    "Forbidden control chars found (ASCII values <32 except 9, 10 and 13)."); `

should print pkg , for us know where this happens , or give any sort of clue how and where this happens

ppisar:
A caller to cr_xml_dump() knows the package name. Or did you mean pointing to the bad byte in the package's textual metadata? The precise location would be great. But we will need to add a new function because cr_Package_contains_forbidden_control_chars() is missing an argument for returning the message and the function is listed in public header files.

sergiomb2:
I mean the text around the bad character

ppisar:
Thanks for the clarification. I recorded the request as issue #424.

sergiomb2:
I propose this PR:  https://github.com/rpm-software-management/createrepo_c/pull/425

Issue #145, opened 2019-04-02
wheels on PyPI


carlwgeorge:
I'm using the createrepo_c Python bindings in a project of mine.  For the CI test run I'm using the [`python:3.7`](https://hub.docker.com/_/python) container image.  I'd like to be able to `pip install createrepo_c`, but this fails due to multiple missing system libraries.  I can work around this by installing them with apt first (at a minimum cmake, libmagic-dev, and librpm-dev), but I believe if a wheel package of createrepo_c was uploaded to PyPI it would just work.

Can a createrepo_c wheel be uploaded to PyPI?

hroncok:
note that the createrepo-c PyPi package appears to be maintained by pulp maintainers.

carlwgeorge:
*/me checks [pulp/pulp_rpm](https://github.com/pulp/pulp_rpm) commit history for recently active maintainers...*

@goosemania @ipanova @daviddavis @dralley Would y'all be able to help here?

dralley:
Hello @carlwgeorge 

Currently uploading a binary wheel of createrepo_c to PyPI is blocked by some changes in the Python build infrastructure https://github.com/pypa/manylinux/pull/279

Since compiled packages link against system libraries, and the system libraries are different on different systems, Python created a build environment called "manylinux" and then built tooling that wraps up all the dynamically-linked libraries into the wheel so that it doesn't depend on any system-provided libraries.  In order to upload a binary wheel to PyPI, it has to have been created in this build environment because otherwise a package built on Ubuntu and uploaded would not work on CentOS.

Unfortunately, that common build environment is based on CentOS 5 and createrepo_c won't build on CentOS 5.  The new environment discussed in the PR above is based on CentOS 6 and that will be usable.  But it's taking a really, really long time for them to get it ready.

carlwgeorge:
Thanks for the clarification @dralley.  I am vaguely familiar with manylinux, but I didn't know that it was *still* building against EL5.  That's understandable that a createrepo_c wheel is blocked on the build environment getting refreshed to EL6.

dralley:
Yup, sorry I couldn't be more helpful!  It was disappointing to us too when we realized that.

carlwgeorge:
Hey, having a clear understanding of the blocker is absolutely helpful.  Thanks!

hroncok:
> but I didn't know that it was still building against EL5

manylinux1 will always build against EL5.

[...]

PR #167, opened 2019-06-13
Add missing python metadata to python2/3-createrepo_c (RhBug:1695677)

kontura:
In order not to depend on scikit-build I added additional python setup
script (setup_for_python_metadata.py) which uses just setuptools and is
used to generate egg-info metadata during building of packages
python2/3-createrepo_c.

Related to #145

### DIFF
--- a/createrepo_c.spec
+++ b/createrepo_c.spec
@@ -251,11 +251,13 @@ ln -sr %{buildroot}%{_bindir}/modifyrepo_c %{buildroot}%{_bindir}/modifyrepo
 %if %{with python2}
 %files -n python2-%{name}
 %{python2_sitearch}/%{name}/
+%{python2_sitearch}/%{name}-%{version}-py%{python2_version}.egg-info
 %endif
 
 %if %{with python3}
 %files -n python3-%{name}
 %{python3_sitearch}/%{name}/
+%{python3_sitearch}/%{name}-%{version}-py%{python3_version}.egg-info
 %endif
 
 %changelog
--- a/src/python/CMakeLists.txt
+++ b/src/python/CMakeLists.txt
@@ -88,4 +88,7 @@ IF (SKBUILD)
 ELSE ()
     INSTALL(FILES createrepo_c/__init__.py DESTINATION ${PYTHON_INSTALL_DIR}/createrepo_c)
     INSTALL(TARGETS _createrepo_c LIBRARY DESTINATION ${PYTHON_INSTALL_DIR}/createrepo_c)
+
+    # Version has to be passed as last argument.
+    INSTALL(CODE "EXECUTE_PROCESS(COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_SOURCE_DIR}/utils/setup_for_python_metadata.py install_egg_info --install-dir \$ENV{DESTDIR}/${PYTHON_INSTALL_DIR} ${VERSION})")
 ENDIF ()
--- a/utils/setup_for_python_metadata.py
+++ b/utils/setup_for_python_metadata.py
@@ -1,0 +1,32 @@
+from distutils.core import setup
+import sys
+
+# This is a simple and fragile way of passing the current version
+# from cmake to setup as I assume no one else will use this.
+#
+# This script has to have the version always specified as last argument.
+version = sys.argv.pop()
+
+setup(
+    name='createrepo_c',
+    description='C implementation of createrepo',
+    version=version,
+    license='GPLv2+',
+    author='RPM Software Management',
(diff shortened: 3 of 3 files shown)


createrepo_c.spec
```diff
@@ -251,11 +251,13 @@ ln -sr %{buildroot}%{_bindir}/modifyrepo_c %{buildroot}%{_bindir}/modifyrepo
 %if %{with python2}
 %files -n python2-%{name}
 %{python2_sitearch}/%{name}/
+%{python2_sitearch}/%{name}-%{version}-py2.7.egg-info/
```
carlwgeorge: You might want to use `%{python2_version}` and `%{python3_version}` instead of `2.7` and `3.7`.  There isn't going to be a 2.8, but python3 is going to continue to increment.

kontura: yes, thank you.

setup_for_python_metadata.py
```diff
@@ -0,0 +1,44 @@
...
+    package_dir={
+        'createrepo_c': 'src/python/createrepo_c'
+    },
+    entry_points={
+        'console_scripts': [
+            'createrepo_c=createrepo_c:createrepo_c',
+            'mergerepo_c=createrepo_c:mergerepo_c',
+            'modifyrepo_c=createrepo_c:modifyrepo_c',
+            'sqliterepo_c=createrepo_c:sqliterepo_c'
+        ]
```
Conan-Kudo: Why are we generating entry points? Won't these overwrite the ones generated by CMake?

src/python/CMakeLists.txt
```diff
@@ -88,4 +88,7 @@ IF (SKBUILD)
 ELSE ()
     INSTALL(FILES createrepo_c/__init__.py DESTINATION ${PYTHON_INSTALL_DIR}/createrepo_c)
     INSTALL(TARGETS _createrepo_c LIBRARY DESTINATION ${PYTHON_INSTALL_DIR}/createrepo_c)
+
+    # Version has to be passed as last argument.
+    INSTALL(CODE "EXECUTE_PROCESS(COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_SOURCE_DIR}/setup_for_python_metadata.py install_egg_info --install-dir \$ENV{DESTDIR}/${PYTHON_INSTALL_DIR} ${VERSION})")
```
Conan-Kudo: Can we move this out of the top level directory so people don't think they can use it normally?

kontura: The utils directory seems like the best fit to me.

setup_for_python_metadata.py
kontura: I just blindly copied them from the original setup.py. Thank you for catching that.

utils/setup_for_python_metadata.py
```diff
@@ -0,0 +1,36 @@
+from setuptools import setup
+import sys
+
+# This is a simple and fragile way of passing the current version
+# from cmake to setup as I assume no one else will use this.
+#
+# This scprit has to have the version always specified as last argument.
```
Conan-Kudo: `scprit` -> `script`

utils/setup_for_python_metadata.py
```diff
@@ -0,0 +1,36 @@
+from setuptools import setup
```
Conan-Kudo: We don't need setuptools for this, we can use `distutils.core`, since we have no setuptools dependency definitions.

Cf. https://docs.python.org/3/distutils/introduction.html

kontura: I have updated it to use only `distutils.core`, this way it generates only one egg-info file instead of a directory so I have also updated the specfile.

utils/setup_for_python_metadata.py
```diff
@@ -0,0 +1,36 @@
...
+        'Programming Language :: Python :: 2.7',
+        'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: 3.4',
+        'Programming Language :: Python :: 3.5',
+        'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.7',
+    ],
+    packages=['createrepo_c'],
+    package_dir={
+        'createrepo_c': 'src/python/createrepo_c'
```
Conan-Kudo: Does this actually work? What is this relative to?

kontura: It seems package_dir and packages attributes are only used when building, installing.. I removed both of them, they don't seem to be needed for generating the egg-info. 

I also checked and all of the other attributes are used in the resulting metadata.

PR #182, opened 2019-10-24
Fix file mode for cache (RhBug:1686812)

kontura:
As was pointed out here: https://github.com/rpm-software-management/createrepo_c/commit/3018bc474f4eab634d176d5150b87ff29751fd65#r35631776 my previous implementation was simply wrong and using `g_mkstemp_full` is better anyway.

https://bugzilla.redhat.com/show_bug.cgi?id=1686812

### DIFF
--- a/src/dumper_thread.c
+++ b/src/dumper_thread.c
@@ -34,6 +34,7 @@
 #include "misc.h"
 #include "parsepkg.h"
 #include "xml_dump.h"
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
@kontura You need to fix your computer date stamps. I just noticed that your commits are dated July 11. I'm dead certain you didn't time-travel. 😜

PR #184, opened 2019-11-25
export all changelog entries in compatibility mode

M0ses:
the old python based createrepo exported all changelog entries to the
other.xml. createrepo_c limits to 10 entries.

This patch is an attempt to get all entries into the other.xml without
changing the default behavior of createrepo_c via using the
--compatibility option.

### DIFF
--- a/src/cmd_parser.c
+++ b/src/cmd_parser.c
@@ -31,7 +31,6 @@
 
 
 #define ERR_DOMAIN                      CREATEREPO_C_ERROR
-#define DEFAULT_CHANGELOG_LIMIT         10
 #define DEFAULT_CHECKSUM                "sha256"
 #define DEFAULT_WORKERS                 5
 #define DEFAULT_UNIQUE_MD_FILENAMES     TRUE
--- a/src/cmd_parser.h
+++ b/src/cmd_parser.h
@@ -24,6 +24,8 @@
 #include "checksum.h"
 #include "compression_wrapper.h"
 
+#define DEFAULT_CHANGELOG_LIMIT         10
+
 
 /**
  * Command line options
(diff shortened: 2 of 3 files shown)


Conan-Kudo:
@M0ses Rather than `--compatibility`, why not just say `--export-all-changelog-entries`?

kontura:
Given that the `--compatibility` option already exists I would prefer using it over creating a new option. But I think we should also describe the change in the option's documentation.

M0ses:
I could also use a new option, but the documentation says:

`Enforce maximal compatibility with classical createrepo`

So what does this mean?
Should their be another option `--be-real-compatible-with-old-createrepo` or something like that?
I think it also may be worth to change the behavior of `--changelog-limit` in compatibility mode.
ATM `--changelog-limit=-1` in createrepo_c gives all entries (N) and the python version uses N-1.

Conan-Kudo:
@M0ses Perhaps that does make sense... Wiring that default into `--compatibility` makes sense, as long as it can still be overridden by `--changelog-limit`.

However, I don't know if it makes sense to implement the old pycreaterepo's changelog limit behavior...

M0ses:
I really have no strong opinion here. 
I thought for the option `--compatibility` should do what is documented.
Feel free to close/reject my PR.

kontura:
I agree @M0ses, I meant that currently the options says: 
`Enforce maximal compatibility with classical createrepo (Affects only: --retain-old-md).`
And with this change I would like if it said something like:
`Enforce maximal compatibility with classical createrepo (Affects: --retain-old-md and default changelog_limit).`

Also if you want to implement the old behavior of `--changelog-limit` with `--compatibility` I would be fine with it because it really is in the documentation.

Issue #180, opened 2019-10-15
Exclude directories as well


xsuchy:
When you run `createrepo_c -x foo*` then it will skip `foo.rpm` but will traverse `fooBAR/foo/foo....` structure.

We have in Copr directories in repo directory which does not contain any rpm packages (e.g., 'configs', 'chroot_scans'). To speed up Copr, I want to instruct createrepo_c to not traverse those directories. This is now not possible.

This can be implemented as part of `-x` or as a new option.

praiskup:
Can we reopen this?  The fix is incomplete or wrong, as it doesn't affect the `--pkglist` files (we still only match filenames, not paths, there).

I'd propose to have `--exclude-dirs` option, without pattern matching.

Conan-Kudo:
This issue is not closed, so I guess it's still up for fixing...

praiskup:
I was wrong.  IMO this can be closed .... because this issue doesn't try to resolve problems with pattern matching, but optimize the directory walk.  And the commit really does what it should....

PR #186, opened 2019-12-02
Skip directory matching --exclude right away.

dturecek:
Fixes issue #180.


Conan-Kudo:
So, this excludes `foo.rpm` in all subdirs?

dturecek:
It excludes `foo` dir and everything inside.

PR #190, opened 2019-12-22
createrepo_c: use pkg href for cache lookup with --update

praiskup:
Duplicated RPM filenames in traversed directory caused cache misses
before (since duplicates are removed from the caches entirely).

Lookup by package location minimizes cache misses a lot.  This required
inventing a new hashing type CR_HT_KEY_HREF.

### DIFF
--- a/src/createrepo_c.c
+++ b/src/createrepo_c.c
@@ -627,7 +627,7 @@ main(int argc, char **argv)
 
     if (task_count && cmd_options->update) {
         int ret;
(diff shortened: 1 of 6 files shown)


praiskup:
We may want to have this configured, dunno.  If this idea is acceptable at all.
Background story comes from [1], where we deal with 28k+ packages, and
about 4k duplicated package filenames.

Conan-Kudo:
This was part of #191, so closing.

PR #191, opened 2019-12-22
recycle pkglist with --update

praiskup:
```
    createrepo_c: add --recycle-pkglist option
    
    This option is useful e.g. for very large repositories in copr [1] where
    we need to update the repodata after each package build or package
    removal.  Re-generating everything from scratch with --update is simply
    too expensive (actually even the directory traversal is) so the only
    only viable option would be to maintain complete list of packages
    manually, and use '--pkglist --no-stat'.
    
    With '--recycle-pkglist' we can avoid the additional work with separate
    pkglists.  Adding new builds can be done just like
      $ createrepo_c --no-stat --update --recycle-pkglist --pkglist <sth>
    and removing something can be done with
      $ createrepo_c --no-stat --update --recycle-pkglist --exclude <sth>
    
    [1] https://pagure.io/copr/copr/pull-request/1150
```

### DIFF
--- a/createrepo_c.bash
+++ b/createrepo_c.bash
@@ -70,7 +70,7 @@ _cr_createrepo()
             --retain-old-md-by-age --cachedir --local-sqlite
             --cut-dirs --location-prefix
             --deltas --oldpackagedirs
-            --num-deltas --max-delta-rpm-size' -- "$2" ) )
+            --num-deltas --max-delta-rpm-size --recycle-pkglist' -- "$2" ) )
     else
         COMPREPLY=( $( compgen -d -- "$2" ) )
     fi
--- a/doc/createrepo_c.8
+++ b/doc/createrepo_c.8
@@ -90,6 +90,14 @@ Specify a text file which contains the complete list of files to include in the 
 .SS \-n \-\-includepkg PACKAGE
 .sp
 Specify pkgs to include on the command line. Takes urls as well as local paths.
+.SS \-\-recycle\-pkglist
+.sp
+Useful only with \fB\-\-update\fR.  Read the list of packages from old metadata,
+and reuse it instead of (perhaps expensive) directory traversal.  This doesn't
+collide with explicitly selected packages by \fB\-\-pkglist\fR or
+\fB\-\-includepkg\fR, such packages are appended to the recycled list.
+This option is useful for I/O optimal repo modifications (package removal by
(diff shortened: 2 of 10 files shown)


praiskup:
This is now tentatively built on top of #190, I need to realize what is acceptable and
what not.  The idea behind this implementation is to not re-load the old metadata
twice, (a) for getting the list of old RPMs, and (b) for --update caches.

As it turned out, the loaded list with  CR_HT_KEY_FILENAME was incomplete, for
the copr PR 1150 case - we lost ~8k filenames because of 4k duplicates.  So I went
with #190 to still keep loading the old metadata only once.

If that is not acceptable, we could indeed load metadata twice...  with CR_HT_KEY_HASH
for the filenames, and second with CR_HT_KEY_FILENAME for caches (as before).

praiskup:
Btw., the normal run of `createrepo --update` takes 6minutes on the sample
I mentioned before, and `createrepo --recycle-pkglist --skip-stat --update` only
like 16s.

Conan-Kudo:
This seems vaguely similar (conceptually) to #73, isn't it?

praiskup:
Similar, but in Copr we actually expect that some files will be dropped from the list,
and some will be added.  So this PR seems to be superset of #73.  I'll take a look
once more a bit later :-)

praiskup:
Looking again, I think that these two PR complement each other .. this PR is to maximize
caching, to avoid expensive directory walk, and allow rather small repo updates.  The other
PR is to not change repodata if no RPM changed, which isn't solved here.

Conan-Kudo:
@kontura Could you please review this and #190 and make a judgement of how we want to proceed here? I'm not sure what we should do here...

praiskup:
Updated according to @kontura's feedback, thanks!  That said, whole metadata loading
(including the logic for `--update-md-path` is moved to separate function now.

Conan-Kudo:
Where's @kontura's feedback? I have no idea what's going on here..

kontura:
Ah sorry we discussed it in person. I basically just said that the new option `--recycle-pkglist` should also take into account the metadata from `--update-md-path`.

And also I will further look into the new key `CR_HT_KEY_HREF` but so far I think there are no drawbacks to this. I will do some more testing and report back.

praiskup:
Awesome!  Thank you.

Conan-Kudo:
@praiskup new release 0.15.5 has your additions...

PR #192, opened 2019-12-29
Sync --excludes matching for dir-walk vs. --pkglist

praiskup:
Make sure that we use patterns the same way for directory walk (we used
to match full relative directory) and for explicit --pkglist (we used to
match only the rpm basename).  So from now on, the relative path is
tested for match.


Conan-Kudo:
How does this relate to #190 and #191?

praiskup:
It doesn't, I realized later.  Sorry for wrong reference.

PR #205, opened 2020-02-13
Fix build on FreeBSD

mikhirev:
Hi!

I got a couple of errors while trying to compile on FreeBSD. First is that `SIGPOLL` is not defined there, and second is that glib libraries are installed to the `/usr/local/lib` directory that was not searched by linker. Please pull fixes for these issues.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,9 +47,13 @@ find_package(ZLIB REQUIRED)
 
 # Add include dirs
 
+include_directories(${BZIP2_INCLUDE_DIRS})
+include_directories(${CURL_INCLUDE_DIRS})
+include_directories(${EXPAT_INCLUDE_DIRS})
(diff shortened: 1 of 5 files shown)


mikhirev:
After some investigation, I replaced `FindGLIB2.cmake` and `FindGTHREAD2.cmake` files. The first one is from the [KDE ECM project][ecm], and the second one is slightly modified first. These modules provide full path to libraries instead `-lglib-2.0` and `-lgthread-2.0`.

I also included missing headers to `src/misc.c` because there were warnings about undefined `WIFEXITED`, `WEXITSTATUS` etc.

[ecm]: https://github.com/KDE/extra-cmake-modules/blob/master/find-modules/FindGLIB2.cmake

Conan-Kudo:
@mikhirev Thanks for the patches! I've cut the new release 0.15.7 which includes these fixes.

PR #207, opened 2020-02-15
Automatically build and upload PyPI packages on release

dralley:
Don't feel obligated to merge this -- we can continue maintaining the package ourselves just fine.

But it would simplify things if the packages were just automatically created and uploaded on every release, especially if we ever start creating binary wheels that would work across different Linux distros (see the discussion here: [0]).

The RPM team or someone from the RPM team would need to be a co-maintainer on the ```createrepo_c``` Python package, and add their credentials to the ```secrets``` storage on this repository [1]

[0] https://github.com/rpm-software-management/createrepo_c/issues/145#issuecomment-583654675
[1] https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#creating-encrypted-secrets


.github/workflows/python_build_and_publish.yml
```diff
@@ -0,0 +1,28 @@
+name: Python package build and publish
+
+on:
+  release:
+    types: [published]
```
dralley: I'm not certain if ```created``` or ```published``` would be better here. The exact distinction is not well documented.

dralley:
If you're OK with this, I'll add it to ```libcomps``` as well.

Conan-Kudo:
@dralley Wouldn't this be a problem since we can't statically link librpm in, among other things?

dralley:
@Conan-Kudo Well, right now it only builds source distributions.  That means you have to have all of the build dependencies installed, but when you install that package the build process is automated away.  It's somewhat limiting but still better than nothing.  

The other form of Python package is a pre-compiled binary "wheel" distribution.  I've been working on this for a good part of the afternoon but [haven't gotten it working yet](https://github.com/dralley/createrepo_c/runs/446874387?check_suite_focus=true#step:7:387).  These packages are effectively statically linked although my understanding is that the actual process is a little bit more complex than that.

For the latter kind of package, once I actually get it to work, it would be entirely standalone.

dralley:
To amend my previous comment a bit:  you can actually build a binary wheel of createrepo_c just fine... but you can't upload it to PyPI unless you've built it a certain way, that accomplishes the "effectively statically linked" part of what I just mentioned.

To do that, you have to build it on a special platform (```manylinux```) that is provided as a docker container, and then run a special tool that takes everything it was dynamically linked against and shoves it into the package, and rewrites the binary to accept it.  This is the part I haven't gotten to work yet.  Hopefully soon.

^ That sounds crazy, but it's the official way of doing things.  It's how ```numpy``` and like packages can work on Fedora, Debian, CentOS, etc...

dralley:
@Conan-Kudo Here's the sole issue blocking this: https://github.com/pypa/auditwheel/issues/229#issuecomment-587943327

I'll keep looking into it

dralley:
So this PR is blocked on a Python ecosystem issue that I (apparently) uncovered: https://github.com/pypa/auditwheel/issues/230

dralley:
So this works now, the only thing missing is the "secrets" for uploading to PyPI.

src/python/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ if (NOT SKBUILD)
 endif (NOT SKBUILD)
 
 EXECUTE_PROCESS(COMMAND ${PYTHON_EXECUTABLE} -c "from sys import stdout; from distutils import sysconfig; stdout.write(sysconfig.get_python_lib(True))" OUTPUT_VARIABLE PYTHON_INSTALL_DIR)
```
dralley, line 30: It's been a while so I don't remember why it was that this was necessary.

dralley:
This issue is still blocking publishing binary packages: https://github.com/pypa/manylinux/issues/823

However, automating publishing the standard "source distributions" as we are currently doing would work fine, if you're interested in merging this.

dralley:
Recreating this PR from a different local branch.

Issue #206, opened 2020-02-15
Fix PyLong_AsSize_t on Python 2


SchweinDeBurg:
```
diff --git a/src/python/updaterecord-py.h b/src/python/updaterecord-py.h                                                                                                
index 1dcea7b..6027a23 100644                                                                                                                                           
--- a/src/python/updaterecord-py.h                                                                                                                                      
+++ b/src/python/updaterecord-py.h                                                                                                                                      
@@ -29,4 +29,14 @@ extern PyTypeObject UpdateRecord_Type;                                                                                                               
 PyObject *Object_FromUpdateRecord(cr_UpdateRecord *rec);
 cr_UpdateRecord *UpdateRecord_FromPyObject(PyObject *o);

+#if PY_MAJOR_VERSION < 3                                                                                                                                               
+#if SIZE_MAX == UINT32_MAX                                                                                                                                             
+#define PyLong_AsSize_t PyLong_AsUnsignedLong                                                                                                                          
+#elif SIZE_MAX == UINT64_MAX                                                                                                                                           
+#define PyLong_AsSize_t PyLong_AsUnsignedLongLong                                                                                                                      
+#else                                                                                                                                                                  
+#error "Unsupported architecture"                                                                                                                                      
+#endif                                                                                                                                                                 
+#endif                                                                                                                                                                 
+                                                                                                                                                                       
 #endif
```

kontura:
I have created a PR https://github.com/rpm-software-management/createrepo_c/pull/208, can you check whether if fixes your problem?

PR #208, opened 2020-02-17
Replace PyLong_AsSize_t with more supported alternative

kontura:
To resolve https://github.com/rpm-software-management/createrepo_c/issues/206


Conan-Kudo:
@kontura Are you sure this doesn't cause problems? Changing datatypes like this makes me a bit nervous...

kontura:
@Conan-Kudo I am pretty sure it should be fine here, its just an internal representation that lives only for a couple of lines. I would even say `unsigned long long` is actually better because it has strictly defined size where as `size_t` can vary.  

It also has enough space for any realistic epoch value.

Issue #150, opened 2019-04-21
RFE: rewrite createrepo_c to use libxml2 only


kloczek:
For some reasons createrepo_c uses two xml parsers and it is not possible to not use one of them
I think that it will be good to rewrite it to use only libxml2 to minimise dependencies.

kontura:
Can you be a bit more specific what exactly do you mean by two xml parsers? When it comes to xml dependencies I have only been able to find libxml2, even in the spec file there is only [this](https://github.com/rpm-software-management/createrepo_c/blob/4570d46b7837d0dbdb7b5f9ffb1e77ea7068c316/createrepo_c.spec#L44).

Or is there some other xml dependency I am missing?

kloczek:
createrepo_c uses expat and libxml2 which are libraries proving parser and some additional set of operation s on xml objects/files.

kontura:
oh I see, you are right.
One more thing I have looked into the commit log where the Expat library was introduced and it seems like it was [deliberately put there to replace libxml2](https://github.com/rpm-software-management/createrepo_c/commit/4c5c0c0b71f588c9303d3d41df6d0687cd06535e) when parsing xml. 

I can't find any specific reason why it was done, but even though it would be better to have just one I think we should first test whether it wasn't for example much slower or something like that.

kloczek:
In the patch there is no any traces about causes of such switch.
Something is wrong with libxml2 locate_and_load_xml_metadata_2()?

kloczek:
AFAIK libxml2 was generally faster than expat.
Memory consumption was as same smaller and this is why generally almost everywhere is used libxml2 and only in few cases expat.

kloczek:
https://cpp.libhunt.com/compare-expat-vs-libxml2

tpgxyz:
```
[tpg@tpg-virtualbox createrepo_c-0.14.0]$ grep -ri libxml | wc -l
33
```
Looks like xml_dump*.{c,h} and repomd.c still requires libxml2

kloczek:
Is it any problem with libxml2 that createrepo_c seems is trying to move away from that library and use expat? (which is now less and less used)

Conan-Kudo:
Unfortunately, I don't know why we were moving to expat, as generally we've been moving to libxml2 in recent years. That was something @Tojaj was doing back then.

I'd probably say we should switch back to libxml2 completely (as we've done for libsolv and other libraries that used expat).

[...]

PR #210, opened 2020-02-24
Parse snippet of repodata (RhBug:1804308) and remove expat

kontura:
Add functions cr_xml_parse_(primary/filelists/other)_snippet and their python counterparts to the API.

These functions allow parsing of repodata passed as a string (instead of a path to file) which doesn't contain a root xml element (eg. `<filelists>`).

There are also tests for the new functionality.

https://bugzilla.redhat.com/show_bug.cgi?id=1804308


src/xml_parser.h
```diff
@@ -171,6 +223,31 @@ int cr_xml_parse_other(const char *path,
...
+ * @param pkgcb          Package callback. (Called when complete package
+ *                       xml chunk is parsed.). Could be NULL if newpkgcb is
+ *                       not NULL.
+ * @param pkgcb_data     User data for the pkgcb.
+ * @param warningcb      Callback for warning messages.
+ * @param warningcb_data User data for the warningcb.
+ * @param err            GError **
+ * @return               cr_Error code.
+ */
+int cr_xml_parse_other_snippet(const char *path,
```
dralley: Same here

src/xml_parser.h
```diff
@@ -123,6 +123,33 @@ int cr_xml_parse_primary(const char *path,
...
+ *                       xml chunk is parsed.). Could be NULL if newpkgcb is
+ *                       not NULL.
+ * @param pkgcb_data     User data for the pkgcb.
+ * @param warningcb      Callback for warning messages.
+ * @param warningcb_data User data for the warningcb.
+ * @param do_files       0 - Ignore file tags in primary.xml.
+ * @param err            GError **
+ * @return               cr_Error code.
+ */
+int cr_xml_parse_primary_snippet(const char *path,
```
dralley: Signature doesn't match the documentation, this should probably be ```xml_string```

src/xml_parser.h
kontura: Thanks, I have fixed these.

kontura:
Since I was working with the xml parser I took the opportunity and switched the expat parser for libxml2.
Resolves: https://github.com/rpm-software-management/createrepo_c/issues/150
I did some tests with big files and the speeds were comparable. It doesn't seem particularly faster but also not slower.

Conan-Kudo:
@kontura I don't know if I'm reading this correctly, but it _seems_ like the unit tests for the Python function for parsing repodata snippets is _before_ the implementation of the function itself? Does it make sense to have them split in that order? Also, does it make sense to split the implementation of the code and tests into separate commits this way, rather than having the function and its tests in the same commit?

Conan-Kudo:
Nope, nevermind, GitHub's rendering of this is just dumb.

dralley:
So I'm not sure if one of the later changes broke something that had previously been working, but it appears that all of the snippet tests are failing.  

Screenshot from 2020-03-11 12-51-40

Screenshot from 2020-03-11 12-57-03

It appears to be trying to fetch the XML string contents though it were a path.  It's not just an issue with the test, I can also confirm this with my own testing.

dralley:
disregard my last comment

PR #216, opened 2020-04-08
Copy updateCollectionModule on assignment to prevent bogus data (RhBug:1821781)

kontura:
When assigned updateCollectionModule goes out of scope it gets freed
because its references drop to 0, this resulted in bogus data because
the memory was still being used.

This patch ensures we pass the updateCollectionModule by value (it gets
copied on assignment) just like everywhere else in the updateinfo api.

Also adds a test for this case.

https://bugzilla.redhat.com/show_bug.cgi?id=1821781

### DIFF
--- a/src/python/updatecollection-py.c
+++ b/src/python/updatecollection-py.c
@@ -293,15 +293,19 @@ set_str(_UpdateCollectionObject *self, PyObject *value, void *member_offset)
 static int
 set_module(_UpdateCollectionObject *self, PyObject *value, void *member_offset)
 {
+    cr_UpdateCollectionModule *orig, *new;
+
     if (check_UpdateCollectionStatus(self))
         return -1;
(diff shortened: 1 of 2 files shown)


goosemania:
@kontura , it works as expected now. Thank you for the fix!

goosemania:
New Z release with this fix would be greatly appreciated!

kontura:
> New Z release with this fix would be greatly appreciated!

I have replied in the [bz](https://bugzilla.redhat.com/show_bug.cgi?id=1821781) since bugzillas are more tied with the release workflow for us.

kontura:
@Conan-Kudo can you please take a look? I can do a new release if you approve these (it would also be good to get this PR: https://github.com/rpm-software-management/createrepo_c/pull/214). :slightly_smiling_face:

Issue #213, opened 2020-03-18
mergerepo_c noarch + multilib in koji bare mode


tkopecek:
As more koji modes are introduced, we've hit a following problem:

When using "bare" mode (--pkgorigins --all), noarch multilib is not activated (it is only triggerred for --koji). I think, that the best option is to add separate option --noarch-multilib, so we don't pollute semantics for other options.

This one should directly solve https://pagure.io/koji/issue/1664. I'm not sure, if it will solve also https://pagure.io/koji/issue/1865, but maybe yes.

Do you've any opinion about this? (The other option is to expan arch list on koji's side, but I think, that all expansion code now lives in mergerepo_c, so it should be better to do it here).

(I can write a patch, if this is ok solution for you).

kontura:
I agree that its better to do it in mergerepo_c, so this solution works for me.

Maybe just the name `--noarch-multilib` seems a bit confusing to me. Does it mean that it will add the noarch to the list of architectures and expand all the possible specified architectures?

tkopecek:
AFAIK, it is about replicating this line: https://github.com/rpm-software-management/createrepo_c/blob/master/src/mergerepo_c.c#L262
But maybe, part of the problem lies also in 
https://github.com/rpm-software-management/createrepo_c/blob/master/src/mergerepo_c.c#L254

So, probably both pieces should be affected by that option. Any idea for better name (as second is not about noarch)?

kontura:
I have further looked at the issues and from https://pagure.io/koji/issue/1865 it seems that koji always passes the `-a` param to mergerepo_c so the https://github.com/rpm-software-management/createrepo_c/blob/master/src/mergerepo_c.c#L262 should never apply?

Also once some arch is passed it always adds the noarch: https://github.com/rpm-software-management/createrepo_c/blob/master/src/mergerepo_c.c#L193.

The way I understand it one issue it about https://github.com/rpm-software-management/createrepo_c/blob/master/src/mergerepo_c.c#L254 like you said, and the second one is https://github.com/rpm-software-management/createrepo_c/blob/master/src/mergerepo_c.c#L147 which is also set based on `options->koji`.

If am right, its still two separate things, but if we always want both maybe something like `--arch-expand`? dunno.

tkopecek:
Yep, it looks it is the right option. @mikem23 ?

mikem23:
[...]

PR #218, opened 2020-04-29
Add --arch-expand option

kontura:
This is a separate option that does a subset of --koji functionality:
For specified --archlist it adds multilib arches and then expands all of
them.

I have also created some tests for our CI stack: https://github.com/rpm-software-management/ci-dnf-stack/pull/821

For issue: https://github.com/rpm-software-management/createrepo_c/issues/213

### DIFF
--- a/doc/mergerepo_c.8
+++ b/doc/mergerepo_c.8
@@ -1,6 +1,6 @@
 .\" Man page generated from reStructuredText.
 .
-.TH MERGEREPO_C  "2019-07-18" "" ""
+.TH MERGEREPO_C  "2020-04-29" "" ""
 .SH NAME
 mergerepo_c \- Merge multiple rpm-md format repositories together
 .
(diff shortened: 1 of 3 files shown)


tkopecek:
LGTM

doc/mergerepo_c.8
```diff
@@ -1,6 +1,6 @@
 .\" Man page generated from reStructuredText.
 .
-.TH MERGEREPO_C  "2019-07-18" "" ""
+.TH MERGEREPO_C  "2020-04-29" "" ""
```
lukash, line 3: What's the date for?

kontura: I think is supposed to be `The date of the last nontrivial change that was made to the man page.` from [here](http://man7.org/linux/man-pages/man7/man-pages.7.html). But we use this [script](https://github.com/rpm-software-management/createrepo_c/blob/master/utils/gen_rst.py) to generate the man pages from GOptionEntry structs in the code and the script just updates the date. 

Its not perfect because there are some more detailed descriptions in the man pages as opposed to the GOptionEntries that the script overrides, I usually just don't include them in the commit but in general the script is helpful.

The script was probably used to or was meant to generate the man pages at rpm build time.

lukash: I see, thanks. So just on a side note if the man page is actually generated then it shouldn't be stored in git, it should be generated by the build system - unless there's a reason, but then that's probably still a bad design that should be fixed in a different way... :sunglasses:

PR #221, opened 2020-05-18
Additional fixes for various issues discovered by static analysis

kontura:
This fixes a couple of problems I have caused when working on createrepo_c.

Discovered by covscans.

### DIFF
--- a/src/python/updaterecord-py.c
+++ b/src/python/updaterecord-py.c
@@ -288,13 +288,17 @@ get_datetime(_UpdateRecordObject *self, void *member_offset)
         memset(dt, 0, sizeof(struct tm));
         res = strptime(str, "%Y-%m-%d", dt);
         if (res == NULL) {
+            g_free(dt);
             // Try to convert the whole string to a number if it passes it's likely in epoch format
             char *t;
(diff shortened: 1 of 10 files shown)


src/python/updaterecord-py.c
```diff
@@ -292,6 +292,7 @@ get_datetime(_UpdateRecordObject *self, void *member_offset)
             char *t;
             long long int epoch = strtoll(str, &t, 10);
             if(*t == '\0') {
+                g_free(dt);
                 return PyLong_FromLongLong(epoch);
             } else {
                 PyErr_SetString(CrErr_Exception, "Invalid date");
```
lukash, line 298: Not related to the PR, but shouldn't you return NULL here? `dt` is empty here right, so the following code at the end of the function will not produce anything valid?

Also, while I'm commenting here, the Error could be more descriptive, as in saying what is the date it's trying to convert (where did it come from?) and actually include the broken date string too?

Tests for handling the error would be ideal... I didn't go looking where the string comes from.

kontura: You are correct, it was the original behavior even before my epoch enhancement, but lets fix it. Also added a commit with one unit test for this invalid case.

PR #225, opened 2020-07-24
Parse xml snippet in smaller parts (RhBug:1859689)

kontura:
If string passed to xmlParseChunk (function froml libxml2) is too big the function errors out, its safer to parse the snippet in
smaller parts.

Also adds a unit test for this case.

https://bugzilla.redhat.com/show_bug.cgi?id=1859689

The third commit is unrelated and fixes an issue of accessing possible NULL pointer to `tmp_err`.


dralley:
I can confirm that this is fixed in the case where I originally discovered it :)  Thanks @kontura!

Issue #40, opened 2015-11-08
CMake scripts can't generate man pages or Python API documentation


Conan-Kudo:
Please add options for the CMake scripts to generate the man pages from source and the Python API documentation. For distributions that strongly prefer "full build from source", this is really important.

PR #226, opened 2020-07-28
Add the section number to the manual pages.

ppentchev:
Hi,

First, thanks a lot for writing and maintaining this implementation of createrepo!

What do you think about this trivial patch that corrects the invocation of the ".TH" (title) macro, adding the missing section 8 parameter? It seems that whatever generated the troff file was aware that there should have been another value there - there are two spaces between the program name and modification date parameters - but it was somehow invoked in such a way as to not place anything there, not even an empty string in troff syntax.

BTW, are these manual pages really generated from a ReST source? If so, where is that source, and how is the generation performed? Or have they been written up once, converted to troff format, and then edited directly in troff format from then on? If it is the latter, may I humbly suggest using the mdoc troff macro set instead for, at least IMHO, a bit more readability? I could help with converting - okay, well, rewriting really - the manual pages into troff mdoc format.

Thanks again for createrepo-c and all the related libraries, and keep up the great work!

G'luck,
Peter

### DIFF
--- a/doc/createrepo_c.8
+++ b/doc/createrepo_c.8
@@ -1,6 +1,6 @@
 .\" Man page generated from reStructuredText.
 .
-.TH CREATEREPO_C  "2020-07-02" "" ""
+.TH CREATEREPO_C 8 "2020-07-02" "" ""
 .SH NAME
 createrepo_c \- Create rpm-md format (xml-rpm-metadata) repository
 .
--- a/doc/mergerepo_c.8
+++ b/doc/mergerepo_c.8
@@ -1,6 +1,6 @@
 .\" Man page generated from reStructuredText.
 .
-.TH MERGEREPO_C  "2020-04-29" "" ""
+.TH MERGEREPO_C 8 "2020-04-29" "" ""
 .SH NAME
 mergerepo_c \- Merge multiple rpm-md format repositories together
 .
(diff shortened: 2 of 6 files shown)


kontura:
The troff files are generated (well in part), you can do: `./utils/gen_manpage.sh src/cmd_parser.c src/mergerepo_c.c src/modifyrepo_c.c src/sqliterepo_c.c doc/` from the checkout dir. The generation is performed by `./utils/gen_manpage.sh` which in turn calls `./utils/gen_rst.py` (if possible I think your current change would fit the best in there) this script then looks at the actual code (passed as arguments `src/cmd_parser.c ..`) and parses the `GOptionEntry` struct.

It is a bit of a mess, because it doesn't allow more advanced formatting and the man page sometimes needs more detailed information. For these reasons when needed I edit the troff files by hand and when regenerating them via the scripts only commit the wanted changes. 

It is bad and I would like to make it more robust and also fix a related issue: https://github.com/rpm-software-management/createrepo_c/issues/40 but I have to say I have basically zero experience with this so I don't know if the solution you suggested would be good and acceptable here plus I am afraid I don't have much time to investigate now.

ppentchev:
Hi,

Thanks for the quick response!

So you're basically saying that I'm kind of blind and did notice what was sitting right there in the utils/ directory... Yeah, sounds like me, sorry!

So here's another trivial change that teaches gen_rst.py to specify the manual page section. For the present, my original commit that adds the sections to the generated files is still needed, since right now the manual pages in doc/ do indeed differ from what would be recreated. What do you think about this combination as a quick fix for the time being?

G'luck,
Peter

Conan-Kudo:
Looks good to me.

Issue #230, opened 2020-10-31
Memory leaks in PyObject_ToStrOrNull()


dralley:
Will re-file in bugzilla, opening here for the markdown formatting.

The way ```PyObject_ToStrOrNull()``` is defined and used fundamentally results in severe memory leakage.

Definition of ```PyObject_ToStrOrNull()``` from ```typeconversions.c```:

```C
char *
PyObject_ToStrOrNull(PyObject *pyobj)
{
    if (PyUnicode_Check(pyobj)) {
        pyobj = PyUnicode_AsUTF8String(pyobj);    // <<--- creates a brand new PyObject [0], reference count never decremented, deallocation never happens
    }

    if (PyBytes_Check(pyobj)) {
        return PyBytes_AsString(pyobj);    // <<--- Returns a pointer to the internal buffer of the leaked object - this only works because the value was leaked and therefore has infinite lifetime
                                           // Reference to bytes object lost forever
    }

    return NULL;
}
```

Results in Valgrind output that looks like this (captured by running the unit tests under valgrind -- under larger-scale use, one could imagine that it could build up to greater impact).

```
==16301== 245 bytes in 7 blocks are definitely lost in loss record 6,793 of 9,219
==16301==    at 0x4C29F73: malloc (vg_replace_malloc.c:309)
==16301==    by 0x4EE796B: PyBytes_FromStringAndSize (in /usr/lib64/libpython3.6m.so.1.0)
==16301==    by 0xF9759D4: PyObject_ToStrOrNull (in /home/vagrant/devel/community/createrepo_c/src/python/createrepo_c/_createrepo_c.so)
==16301==    by 0xF975B90: PyObject_ToDependency (in /home/vagrant/devel/community/createrepo_c/src/python/createrepo_c/_createrepo_c.so)
==16301==    by 0xF973B4D: set_list (in /home/vagrant/devel/community/createrepo_c/src/python/createrepo_c/_createrepo_c.so)
==16301==    by 0x4F1DD5E: _PyObject_GenericSetAttrWithDict (in /usr/lib64/libpython3.6m.so.1.0)
==16301==    by 0x4F1DEFC: PyObject_SetAttr (in /usr/lib64/libpython3.6m.so.1.0)
==16301==    by 0x4F7CA5D: _PyEval_EvalFrameDefault (in /usr/lib64/libpython3.6m.so.1.0)
[...]

PR #231, opened 2020-11-01
Fix memory leaks in python extension code

dralley:
fixes: #230 (bz [#1893459](https://bugzilla.redhat.com/show_bug.cgi?id=1893459))

Additionally fixes a couple of other memory leaks.  Completely removes the ```PyObject_ToStrOrNull()``` function which cannot possibly work without leaking memory.


goosemania:
Thanks, @dralley for findingthe spot and for your fix!

Just to give an idea of the level of improvement it brings. 
Processing 300 000 RPMs, went from using 20GB+ of RAM to 1GB :open_mouth:   :upside_down_face:

kontura:
That's amazing.

Thank you!

Issue #234, opened 2020-11-09
Using parts of the Python API with Python > 3.7 prints DeprecationWarning


dralley:
On Fedora 32 with Python 3.8, running the Python tests will print the following warnings.  These were introduced by https://bugs.python.org/issue36381 to warn about an upcoming Python C API change, which is explained further [here](https://github.com/python/cpython/commit/d3c72a223a5f771f964fc34557c55eb5bfa0f5a0).

```
tests/python/tests/test_contentstat.py::TestCaseContentStat::test_contentstat_ref_in_crfile
  /home/dalley/devel/community/createrepo_c/tests/python/tests/test_contentstat.py:83: DeprecationWarning: PY_SSIZE_T_CLEAN will be required for '#' formats
    f.write("foobar")

tests/python/tests/test_crfile.py::TestCaseCrFile::test_crfile_bz2_compression
  /home/dalley/devel/community/createrepo_c/tests/python/tests/test_crfile.py:88: DeprecationWarning: PY_SSIZE_T_CLEAN will be required for '#' formats
    f.write("foobar")

tests/python/tests/test_crfile.py::TestCaseCrFile::test_crfile_gz_compression
  /home/dalley/devel/community/createrepo_c/tests/python/tests/test_crfile.py:76: DeprecationWarning: PY_SSIZE_T_CLEAN will be required for '#' formats
    f.write("foobar")

tests/python/tests/test_crfile.py::TestCaseCrFile::test_crfile_no_compression
  /home/dalley/devel/community/createrepo_c/tests/python/tests/test_crfile.py:65: DeprecationWarning: PY_SSIZE_T_CLEAN will be required for '#' formats
    f.write("foobar")

tests/python/tests/test_crfile.py::TestCaseCrFile::test_crfile_operations_on_closed_file
  /usr/lib64/python3.8/unittest/case.py:202: DeprecationWarning: PY_SSIZE_T_CLEAN will be required for '#' formats
    callable_obj(*args, **kwargs)

tests/python/tests/test_crfile.py::TestCaseCrFile::test_crfile_xz_compression
  /home/dalley/devel/community/createrepo_c/tests/python/tests/test_crfile.py:100: DeprecationWarning: PY_SSIZE_T_CLEAN will be required for '#' formats
    f.write("foobar")

tests/python/tests/test_crfile.py::TestCaseCrFile::test_crfile_zck_compression
[...]

PR #235, opened 2020-11-09
Fix Python deprication (PY_SSIZE_T_CLEAN)

kontura:
PyArg_ParseTuple() and formats which use # require the PY_SSIZE_T_CLEAN
macro since python 3.10

For: https://github.com/rpm-software-management/createrepo_c/issues/234
This should also fix: https://bugzilla.redhat.com/show_bug.cgi?id=1891785

### DIFF
--- a/src/python/compression_wrapper-py.c
+++ b/src/python/compression_wrapper-py.c
@@ -17,6 +17,7 @@
  * USA.
  */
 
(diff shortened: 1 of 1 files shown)


m-blaha (approved):
LGTM

PR #238, opened 2020-12-15
Never leave behind .repodata lock on exit (RhBug:1906831

kontura:
createrepo_c was removing `.repodata` if `exit_val` was se to failure.
Problem is `exit_val` is set only on a couple of places so most of the
failures (no matter how rare) leave `.repodata` behind.

However because on a successful run `.repodata` is renamed to normal
output `repodata` we know that if `.repodata` are present there was an
error and we can delete them.

https://bugzilla.redhat.com/show_bug.cgi?id=1906831

### DIFF
--- a/src/createrepo_c.c
+++ b/src/createrepo_c.c
@@ -665,9 +665,6 @@ main(int argc, char **argv)
         exit(EXIT_FAILURE);
     }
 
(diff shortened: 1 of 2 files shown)


kontura:
Test: https://github.com/rpm-software-management/ci-dnf-stack/pull/926

PR #243, opened 2021-01-12
Disable drpm support by default (RhBug:1914828)

kontura:
Build with drpm only if specified.

https://bugzilla.redhat.com/show_bug.cgi?id=1914828

Disable delta rpms tests on fedora: https://github.com/rpm-software-management/ci-dnf-stack/pull/934

### DIFF
--- a/createrepo_c.spec
+++ b/createrepo_c.spec
@@ -11,9 +11,13 @@
 
 %if 0%{?rhel} && 0%{?rhel} <= 7
 %bcond_with python3
-%bcond_with drpm
 %else
 %bcond_without python3
+%endif
+
+%if 0%{?rhel} && ( 0%{?rhel} <= 7 || 0%{?rhel} >= 9 )
+%bcond_with drpm
+%else
 %bcond_without drpm
 %endif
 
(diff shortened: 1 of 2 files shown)


createrepo_c.spec
```diff
@@ -11,12 +11,12 @@
 
 %if 0%{?rhel} && 0%{?rhel} <= 7
 %bcond_with python3
-%bcond_with drpm
 %else
 %bcond_without python3
-%bcond_without drpm
 %endif
 
+%bcond_with drpm
```
dmach, line 18: I don't think we want to turn drpm off on Fedora.
This is probably ok on RHEL as long as we want to maintain a downstream patch on RHEL 8 (drpm is enabled there).

Maybe the right thing to do is extending the condition on line 12:
`%if 0%{?rhel} && ( 0%{?rhel} <= 7 || 0%{?rhel} >= 9 )`

kontura: Ok, I have added the condition. :+1:

dmach:
I've reviewed the patch and tested a scratch build in koji with this spec.
Merging.

PR #250, opened 2021-02-25
find Python 3 and libs on CentOS 7 too

evgeni:
CentOS 7 has an old CMake (2.8.12) which is really not smart about finding Python interpreters.

If we pass `EXACT` to `FIND_PACKAGE`, we end up with the following error:

```
  Could NOT find PythonInterp: Found unsuitable version "3.6.8", but required
  is exact version "3" (found /usr/bin/python3)
```

Which is *kinda* true, `3.6.8 != 3`. But in reality, we want *a* Python 3, so let's just not ask for an `EXACT` match here.
The same does not fail on newer CMakes which understand that we only care about the major version of Python here.

But after adjusting that, we get the following:

```
-- Could NOT find PythonLibs (missing:  PYTHON_LIBRARIES PYTHON_INCLUDE_DIRS) (Required is at least version "3")
-- Found PythonInterp: /usr/bin/python3 (found suitable version "3.6.8", minimum required is "3") 
…
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
PYTHON_INCLUDE_DIR (ADVANCED)
   used as include directory in directory /builddir/build/BUILD/createrepo_c-0.17.0/src/python
PYTHON_LIBRARY (ADVANCED)
    linked by target "_createrepo_c" in directory /builddir/build/BUILD/createrepo_c-0.17.0/src/python
```

This can be avoided by *first* searching for the binary, and *then* for the libraries.

I also marked the libraries as `REQUIRED` as we really need them :)


kontura:
It does help, thanks. :+1:

evgeni:
Thanks for merging so quickly!

Are you planning to do a 0.17.1 soon, or shall I apply that patch locally in our environment in the meantime?

kontura:
I can make 0.17.1 soon, how fast do you need it? I would like to wait for my other PR https://github.com/rpm-software-management/createrepo_c/pull/249.

evgeni:
ideally this week, the sooner the better. but really no need to hurry just because of us, we *can* just pick that one change into our packaging.

kontura:
ok, I did the 0.17.1 release now, it could have become an issue for someone else as well so this is safer. :slightly_smiling_face:

evgeni:
Thanks a ton!

Issue #82, opened 2017-10-04
RFE: zstd support


Conan-Kudo:
Please consider adding support for compressing metadata with zstd, in addition to gzip, bzip2, and xz. As RPM 4.14 supports zstd compression of package payloads, it'd be great if we can also produce zstd-compressed metadata, too.

According to [the zstd website](http://facebook.github.io/zstd/), it can offer slightly higher compression ratios with faster compression and decompression times, making it a great choice for metadata. It also can be tuned with a dictionary to make compression even better while being quite fast, which would help with making metadata easier to fetch and use.

dralley:
I'm looking into the kind of benefit that could be expected from this, and it looks very promising. 

I downloaded some metadata from https://dl.fedoraproject.org/pub/fedora/linux/releases/37/Everything/x86_64/os/, decompressed it, and then used zstd benchmark mode on the XML with a range of levels 3 to 10 (beyond that was too costly for the small improvement)

zstd

```
[dalley@thinkpad repodata]$ zstd -b3 -e19 54bbae6e9d4cd4865a55f7558daef86574cddc5f2a4f8a0d9c74f946e1a45dd3-primary.xml
 3#45dd3-primary.xml : 162151977 ->  17209222 (x9.422),  536.9 MB/s, 2212.8 MB/s.xml...       
 4#45dd3-primary.xml : 162151977 ->  17223649 (x9.414),  513.6 MB/s, 2227.1 MB/s.xml...       
 5#45dd3-primary.xml : 162151977 ->  15650479 (x10.36),  229.3 MB/s, 2402.3 MB/s.xml...       
 6#45dd3-primary.xml : 162151977 ->  15100891 (x10.74),  178.1 MB/s, 2318.3 MB/s.xml...       
 7#45dd3-primary.xml : 162151977 ->  14848686 (x10.92),  130.4 MB/s, 2162.0 MB/s.xml...       
 8#45dd3-primary.xml : 162151977 ->  14554603 (x11.14),  106.5 MB/s, 2280.5 MB/s.xml...       
 9#45dd3-primary.xml : 162151977 ->  14548575 (x11.15),   91.2 MB/s, 2220.9 MB/s.xml...       
10#45dd3-primary.xml : 162151977 ->  14391159 (x11.27),   66.7 MB/s, 2234.5 MB/s.xml...
[...]

Issue #53, opened 2016-03-14
RFE: Multi-threaded (de)compression of metadata


Conan-Kudo:
While Mageia was working on implementing usage of createrepo_c to regularly generate rpm-md data, it was discovered that it was much slower on compressing xz metadata than gzip. Looking through the code on compression, it doesn't appear that you're using multi-threaded compression.

As of xz 5.2.0, the multi-threaded API is now stable and can be used in production. The documentation for this was incorrect in 5.2.0 (indicating only decompression support) but that was fixed in 5.2.2.

You may want to implement threaded (de)compression for bzip2 and gzip too.

See [the Mageia bug report](https://bugs.mageia.org/show_bug.cgi?id=17057) for more information.

Tojaj:
Createrepo_c does a lot of things in threads (almost everything that could be reasonably parallelized is parallelized) [1] [2] [3]. RPM headers are read and dumped (generation of xml chunks) in threads, checksums are calculated in threads, repodata files are written and compressed simultaneously in threads (one thread per file) also deltarpms are generated in threads.

I looked on this and switched lzma (xz) encoder from `lzma_easy_encoder` to threaded `lzma_stream_encoder_mt` [4], here are the results:
https://gist.github.com/Tojaj/cfa35863d481e5f3428d
The performance is worse. This has been just a quick and dirty benchmark, I'll take a look on this in more detail soon (I'll try to tweak configuration a bit) but for now, as expected, it seems that adding more threads to already parallelized tasks doesn't bring any performance boost but brings overhead which causes performance degradation.

[1] https://github.com/rpm-software-management/createrepo_c/blob/master/src/dumper_thread.c#L298
[2] https://github.com/rpm-software-management/createrepo_c/blob/master/src/threads.h
[3] https://github.com/rpm-software-management/createrepo_c/blob/master/src/deltarpms.c#L302
[4] http://git.tukaani.org/?p=xz.git;a=blob_plain;f=doc/examples/04_compress_easy_mt.c;hb=HEAD

pterjan:
Thank you for the patch, I just tried it on my 4-core machine:

With this patch: real 1m22.385s
Without this patch: real 1m35.695s

So indeed the improvement is not great.

I then tried on the Mageia upload server (32 cores) and this is a different story:

With this patch: real 0m50.894s
Without this patch: real 1m45.635s

Tojaj:
[...]

PR #251, opened 2021-03-02
Use rpmio (add support for zstd)

kontura:
This adds support for rpmio compression modes, it is possible to use `--general-compress-type 9.zstd` , `9T4.zstd` or `7T16.xzdio` for parallel (de)compression with createrepo_c.

It does break the ABI of the C library because `cr_CompressionType` is no longer an enum but instead a `const char*` and all its values `CR_CW_*_COMPRESSION` become defined as `const char *`. However I believe the API is preserved and anything linking with the C library should work after a recompile. This change is required in order to encompass all the possible rpmio modes.

Should help with: https://github.com/rpm-software-management/createrepo_c/issues/53. https://github.com/rpm-software-management/createrepo_c/issues/82

The plan is to keep both rpmio and the current compression methods for a while and after some transition period keep just rpmio and zchunk. This would allow to remove quite a lot of code.


Conan-Kudo:
I spoke to @dmach about this a while back, and I'm not sure if we want to do this. One of the major reasons for createrepo_c's somewhat weird way of handling RPMs (internal definitions of some tags, internal handling for compression algorithms and payload data, etc.) is to avoid the problem where RPM needs to be upgraded or have backports for createrepo_c to be useful for newer distribution targets. This has saved our bacon in Fedora and openSUSE enough times that I'm not sure I would want to accept changing to use librpmio for this. I can easily foresee this biting us again later as RPM continues to change and people are required to use older distributions for production while wanting to support newer distributions over time.

dralley:
I have to agree with @Conan-Kudo.  Pulp / Satellite would inevitably run into the same problems - any significant changes to the metadata may result in installations on older RHEL (or any other OS) being unable to publish the metadata needed by newer RHEL (or any other OS).  The support timeframes make it a very difficult problem.

lukash:
I may not be fully understanding the problem, so bear with me:

From what I can see we can use any compression functions that are present in rpm-libs (as long as they are present in rpm-libs versions on all systems we build createrepo_c for). If we need to add support for a new compression that is not present in rpm-libs on older systems for which we build createrepo_c, we'd need to add our own implementation (or dependency). So we should keep the code in such a way it's possible to add different compression implementations other than rpm, but that's all? Am I missing something?

dralley:
That's the crux of it, yes

kontura:
I couldn't get this to work properly because:
- we cannot be completely transparent to the compression since createrepo needs to know appropriate compression suffix but there is not way to get it from rpmio for any arbitrary rpmio compression mode
- there is no straightforward way to validate the rpmio compression modes and when an invalid mode is specified by the user rpmio creates the file normally but doesn't use any compression

Based on this and the concerns raised above I have reduced this PR to just adding zstd support through rpmio. That should make things simpler hopefully.

kontura:
Also created some ci tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1029

src/cmd_parser.c
```diff
@@ -156,7 +156,7 @@ static GOptionEntry cmd_entries[] =
     { "xz", 0, 0, G_OPTION_ARG_NONE, &(_cmd_options.xz_compression),
       "Use xz for repodata compression.", NULL },
     { "compress-type", 0, 0, G_OPTION_ARG_STRING, &(_cmd_options.compress_type),
-      "Which compression type to use.", "COMPRESSION_TYPE" },
+      "Which compression type to use. Supported compressions are: bzip2, gzip, zck, zstd, xz.", "COMPRESSION_TYPE" },
```
dralley, line 159: It would be great if "none" were also an option

Conan-Kudo:
I'd rather not have rpmio used for compression at all.

dralley:
As long as it's supported on RHEL8, we can probably live with it.  It will cause problems for the binary Python packages though, which is unfortunate.  I do agree about preferring not to rely on rpmio but we can see what problems it causes for us in practice and perhaps update the implementation later.

Conan-Kudo:
The problem is that it's not supported on RHEL 7, and that's where the majority of Foreman+Pulp/Satellite deployments are.

dralley:
It partially depends on how fast the uptake is.  If it's not actually *used* for another 2 years then it's less of a problem than if third parties and Fedora start using it immediately (requisite on having a new-enough createrepo_c available to begin with).

But you're right that such assumptions are probably not a good idea.

Conan-Kudo:
It was a pain working through the crap caused by RHEL stripping out zstd support when forking from Fedora, and then needing to put it back afterward when Fedora wanted to use zstd for RPM payloads. If I don't have to, I don't want to take on an rpmio dependency for this because I definitely don't want this problem again.

kontura:
I see, ok then I am closing this.

dralley:
https://github.com/rpm-software-management/createrepo_c/issues/145#issuecomment-901461469

dralley:
@kontura @Conan-Kudo For what it's worth, as of the most recent releases of Foreman / Satellite only support EL8+.  I'm not sure if that changes anything for you.

Conan-Kudo:
I don't think it'd be a good idea for us, because it creates unusual constraints for things like Koji and OBS servers too. They are often deployed on the oldest platforms even when builds are being made more newer ones on those systems.

dralley:
Ack.

PR #262, opened 2021-04-21
CI: Add pip to packages Python CI installs

m-blaha:
python3-pip is just a weak dependence of python3. In case the python3
was installed without weakdeps in base container and did not get
upgraded during `dnf builddep`, the pip command is missing.

### DIFF
--- a/.github/workflows/ci-python.yml
+++ b/.github/workflows/ci-python.yml
@@ -12,7 +12,7 @@ jobs:
 
       - name: Install dependencies
         run: |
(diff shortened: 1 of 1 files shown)


kontura:
Oh so this was working before because the base container image had older python which got upgraded during `dnf builddep` (this installed `pip`) but now there is a new base container image with newer python and it doesn't get upgraded -> no `pip`.

I see, thanks for figuring it out. LGTM.

PR #265, opened 2021-04-28
Fix conditionals in spec

pkratoch:
Enable libmodulemd in RHEL >= 8 (RhBug:1816753)

Fix case when 0%{?fedora} or 0%{?rhel} is undefined. For example, on
RHEL, 0%{?fedora} is undefined and "0%{?fedora} < 29" resolves to true,
although you'd expect it to be false.

### DIFF
--- a/createrepo_c.spec
+++ b/createrepo_c.spec
@@ -10,13 +10,13 @@
 %bcond_without drpm
 %endif
 
-%if 0%{?rhel} || 0%{?fedora} < 29
+%if 0%{?rhel}
 %bcond_with zchunk
 %else
 %bcond_without zchunk
 %endif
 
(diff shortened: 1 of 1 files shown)


createrepo_c.spec
```diff
@@ -10,13 +10,13 @@
 %bcond_without drpm
 %endif
 
-%if 0%{?rhel} || 0%{?fedora} < 29
+%if 0%{?rhel} || (0%{?fedora} && 0%{?fedora} < 29)
```
m-blaha, line 13: Does it even make sense to care about so old Fedora? I mean
```
%if 0%{?rhel}
```
might be enough.

pkratoch: Good point, removed.

PR #269, opened 2021-05-24
Add option --skip-filelists

malmond77:
This is a site-local optimization. Some packages and repos include an
enormous number of files. This is extremely expensive if said repo is
also fast changing.

Impact of skipping filelists: breaking ability to resolve file/path
based dependencies, `-f` (file ownership) and `-l` (list) options in
repoquery.


malmond77:
Have pulled this patch into CentOS Stream 8 Hyperscale SIG (as experimental): https://git.centos.org/rpms/createrepo_c/diff/c8s..c8s-sig-hyperscale-experimental and have published: http://mirror.centos.org/centos/8-stream/hyperscale/x86_64/packages-experimental/Packages/c/

dralley:
This seems problematic for dependency solving?  SAT-based dependency solvers like libsolv (used by DNF, Zypper) have a very difficult time with incomplete solutions - how robust are those package managers in terms of handling the incomplete dependencies that this will generate?

(I'd like details on this regardless because Pulp, another libsolv user, handles this situation very very badly, and we would love to do better if that's possible)

@Conan-Kudo

Conan-Kudo:
@dralley is absolutely correct, unfortunately...

dralley:
The only way this *might* be able to work if there was enforcement that any files that are actually depended on are listed in `primary.xml`, and are correctly parsed from `primary.xml` in this mode.

But enforcing that would be a huge problem.  Possibly intractable for createrepo_c since repositories can be (and nearly always are) layered on top of each other, e.g. Appstream packages depending on BaseOS packages, or fedora33-updates packages depending on fedora33-release packages.  It would be impossible to guarantee non-broken repositories without looking at the entire picture.

davide125:
To be clear, the usecase for this on the Facebook side is for an internal repo that has both extremely high churn, and no meaningful usage of dependencies by design. Fully agree that this would not be suitable for standard distro repos.

dralley:
What sorts of before / after numbers are you seeing with this change?

malmond77:
The impact we've seen:

* `*filelists.xml.gz` reduced from 50MiB to 764MiB (67x or 1.5% of original). This affects network I/O and https mirror capacity, and disk writes. This reduces the CPU on the client for decompressing and parsing the xml reflected in the next item.
* Wall clock time of `dnf makecache --refresh` (used once per chef run to obtain a up-to-date perspective) from 22s to 2s.

@dralley, @Conan-Kudo : you're right: this isn't an attempt to make the filelists smaller because it's an intractable problem to generalize. That is why I put this under the "expert" option. This very simple change has had an extraordinary effect for us.

lukash:
I think the documentation of this option should be expanded in detail with regards to the impacts described above, but having the option seems fine as it obviously can have use cases for special repos. Maintainers of such repos should know what they're doing when using this option...

Conan-Kudo:
This is the only risky option we have that actually has lasting impact, I don't know if we want to accept that without some extra guards in place when the option is used. Or at least a warning printed out on stdout and stderr when it's used to indicate the repository is technically broken.

Conan-Kudo:
Note that there has been an ask to make DNF fetch filelists (and regenerate solver cache) on demand only when dependency request needs something that's not in primary.xml, but it just simply hasn't been implemented so far...

malmond77:
I'll expand the docs/help to explain when/why this helps. I'll add stderr logging (or whatever the predominant pattern is) to help reinforce that the repo does not provide filelists which can break file based dependency resolution and repoquery.

doc/createrepo_c.8
```diff
@@ -213,5 +213,8 @@ Exit with retval 2 if there were any errors during processing
 .SS \-\-ignore\-lock
 .sp
 Expert (risky) option: Ignore an existing .repodata/. (Remove the existing .repodata/ and create an empty new one to serve as a lock for other createrepo instances. For the repodata generation, a different temporary dir with the name in format .repodata.time.microseconds.pid/ will be used). NOTE: Use this option on your own risk! If two createrepos run simultaneously, then the state of the generated metadata is not guaranteed \- it can be inconsistent and wrong.
+.SS \-\-skip\-filelists
+.sp
+Expert (risky) option: Skip filelist generation, potentially saving significant bandwidth for repos with large numbers of files in packages. NOTE: Use this option on your own risk! This is a site-local optimization and should not be used for public repos. The site operator warrants the filenames in the packages in the repo are not named as required in other packages. The site operator also warrants that all clients do not need to use repoquery -l and -f to list or find packages that own a given file.
 .\" Generated by docutils manpage writer.
```
dralley, line 219: I would reword this slightly.

>Expert (dangerous) option: Skip filelist generation, potentially saving significant bandwidth for repos with large numbers of files in packages. NOTE: Use this option on your own risk! This is a site-local optimization and should not be used for public repos. This option is known to break dependency resolution for any packages which depend on files provided by packages within this repository, e.g. /bin/bash). It will also prevent using repoquery -l and -f from listing files owned by a package or discover which package owns a particular file.

dralley:
@malmond77 What do you think about @Conan-Kudo's suggestion regarding improving DNF?

>Note that there has been an ask to make DNF fetch filelists (and regenerate solver cache) on demand only when dependency request needs something that's not in primary.xml, but it just simply hasn't been implemented so far...

If the primary concern is bandwidth, this seems like a better long-term approach.

malmond77:
I tried to address the comment about improving DNF above. The problem is intractable because the only way to resolve file based dependencies is to actually get the filelists from all repos. We can't even assume that the repo containing the package that needs a file is the one providing it, or the best choice. I will bring this up at the DNF community meeting this Friday.

Conan-Kudo:
We discussed this in the DNF community meeting and we've agreed to not merge this, because it's too easy to misuse and break things. Instead, DNF will expose an option to support disabling fetching and using filelists selectively and eventually do it on-demand.

davide125:
For future reference, this is now implemented in DNF 5: https://github.com/rpm-software-management/dnf5/commit/96b640a83924796943cbe603569315d91a164114

j-mracek:
The auto-detection was merged but I am not sure whether it is a good idea anyway. At least the autodetection must be improved, because to install using path from /usr/bin I do not need filelists.

PR #270, opened 2021-05-28
Fix typos

dralley:


src/xml_parser_internal.h
```diff
@@ -117,9 +117,9 @@ typedef struct _cr_ParserData {
 
     int do_files;   /*!<
         If == 0 then parser will ignore files elements in the primary.xml.
-        This is useful when you are inteding parse primary.xml as well as
+        This is useful when you are intending parse to primary.xml as well as
```
kontura, line 120: I think this should be: `intending to parse primary.xml`?

Also I would add the `to` after the second `intending` fix as well.

kontura:
Thank you!

PR #271, opened 2021-05-28
Add a high-level API for creating repodata

dralley:
This is a mostly-functional prototype of a more convenient (Python) API for writing repos that I threw together.

There's no obligation to merge this, but, is it something you would be interested in merging?  I figure that 90% of this code is code that any consumer of the API would need to write / copy anyway (and I started with just the example code).

If the answer is yes, I can finish it off, write documentation and tests, etc.  Otherwise we can just close.


dralley:
Not sure why that test would be failing.

kontura:
yea the failing tests are unrelated, in addition to the libxml2 issue you described there is also a zchunk regression.

Regarding the PR it self, I personally would be fine with merging this. Some kind of a higher-level API that is in spirit closer to the createrepo_c binary seems nice to me.

Conan-Kudo:
I really like the idea of a simpler high level API too.

dralley:
Great, I'll keep working on it then.

examples/python/simple_createrepo.py
```diff
@@ -6,93 +6,53 @@
...
-    oth_xml_path = os.path.join(repodata_path, "other.xml.gz")
-    pri_db_path  = os.path.join(repodata_path, "primary.sqlite")
-    fil_db_path  = os.path.join(repodata_path, "filelists.sqlite")
-    oth_db_path  = os.path.join(repodata_path, "other.sqlite")
+    # create a RepositoryWriter with a context manager - finish() is called automatically
+    with cr.RepositoryWriter(path, with_sqlite=True) as writer:
+        writer.set_num_of_pkgs(len(pkg_list))
+        for filename in pkg_list:
+            pkg = writer.add_pkg_from_file(filename)
+            print("Added: %s" % pkg.nevra())
```
dralley, line 23: ~80 lines of code to ~5 :)

dralley:
I have some API design questions, see the TODO notes

src/python/createrepo_c/__init__.py
```diff
@@ -358,6 +362,273 @@ def decompress_file(src, dst, comtype, stat=None):
...
+        os.makedirs(self.path, exist_ok=True)
+
+        # TODO: move this to finish() so that the existing repodata isn't moved until the new metadata has been
+        # successfully written?
+        # TODO: update option - copy over existing tags, packages to the new metadata
+        if self.repodata_path.exists():
+            if preserve_existing_metadata:
+                x = 0
+                while True:
+                    new_repodata_path = f"{self.repodata_path}_{x}"
```
kontura, line 420: I think `createrepo_c` binary and `RepositoryWriter` should be as consistent as possible. How about using something closer to `--retain-old-md` and `--keep-all-metadata` instead of `preserve_existing_metadata`? 

Or do you prefer this behavior?

src/python/createrepo_c/__init__.py
```diff
@@ -358,6 +362,273 @@ def decompress_file(src, dst, comtype, stat=None):
...
+                record.fill(self._metadata_checksum_type)
+                records.append(record)
+
+                if sqlite_name and self._with_sqlite:
+                    sqlite_metadata_info = self.working_metadata_files[sqlite_name]
+                    sqlite_metadata_info.writer.dbinfo_update(record.checksum)
+                    sqlite_metadata_info.writer.close()
+            else:
+                record = RepomdRecord(record_name, str(metadata_info.path))
+                record = record.compress_and_fill(self._metadata_checksum_type, BZ2_COMPRESSION)
```
kontura, line 604: While trying out the `RepositoryWriter` I noticed the created repositories are invalid. I think this section specifically exposes a bug in the library. 

`record.compress_and_fill(self._metadata_checksum_type, BZ2_COMPRESSION)` returns a new record with filled in checksums while the old one (created on the line above - 603) is freed. However these two lines: https://github.com/rpm-software-management/createrepo_c/blob/master/src/repomd.c#L557-L558 make the new record use chunk from the old (freed) one. This leads to garbage values.
Could this be fixed as a part of this PR?

src/python/createrepo_c/__init__.py
dralley: This is just what the example had been doing when I copied it over.  I think there's a good argument to be made either way.  It does seem a little silly to keep the metadata files around but throw away the repomd.xml that tied them together.  If you keep the old repodata directories, it's fairly straightforwards to turn it back into a functional repo (assuming the packages are still there)

src/python/createrepo_c/__init__.py
dralley: @kontura It might take a while before this PR is ready depending on how much time I can dedicate to it.  It should probably go in a separate PR.  I can maybe take a shot at fixing it next week though.

kontura: Ok, that is no problem.

I think we just need to change `record->chunk` to `crecord->chunk` for both of those lines.  :slightly_smiling_face:

dralley:
@kontura Should I go forwards with this, are you good with commits 2 and 3?

kontura:
Yes and maybe we could even merge `metadata_compression` with `general_compression`?

dralley:
This is ready to be looked at.  I don't claim it's perfect but I tried to get decent test coverage.

src/python/createrepo_c/__init__.py
```diff
@@ -382,6 +386,364 @@ def __init__(self, primary_path, filelists_path, other_path, newpkgcb=None, warn
...
+    def add_pkg(self, pkg):
+        """Add a package to the repo from a pre-created Package object."""
+        assert self._has_set_num_pkgs, "Must set the number of packages before adding packages"
+        assert not self._finished, self._FINISHED_ERR_MSG
+
+        self.working_metadata_files["primary"].writer.add_pkg(pkg)
+        self.working_metadata_files["filelists"].writer.add_pkg(pkg)
+        self.working_metadata_files["other"].writer.add_pkg(pkg)
+
+    def add_repomd_metadata(self, name, path, use_compression=True):
```
dralley, line 674: Unsure if `use_compression=True` is a good default

src/python/createrepo_c/__init__.py
```diff
@@ -382,6 +386,364 @@ def __init__(self, primary_path, filelists_path, other_path, newpkgcb=None, warn
...
+            cr.PackageIterator object.
+        """
+        return PackageIterator(
+            primary_path=self._primary_xml_path,
+            filelists_path=self._filelists_xml_path,
+            other_path=self._other_xml_path,
+            warningcb=warningcb,
+        )
+
+    def parse_packages(self, only_primary=False):
```
dralley, line 473: I occasionally find this useful, but I'm not sure if it's useful enough to add to the API.  Thoughts?

dralley:
@kontura ^

dralley:
Bump!

kontura:
@dralley sorry I am a bit swamped now, it might take a while.

src/python/createrepo_c/__init__.py
Conan-Kudo: I think this is quite useful, especially for repodata introspection.

src/python/createrepo_c/__init__.py
Conan-Kudo: Yes, it is. It's what we expect for network transmitted stuff.

Conan-Kudo (approved):

src/python/createrepo_c/__init__.py
```diff
@@ -382,6 +386,364 @@ def __init__(self, primary_path, filelists_path, other_path, newpkgcb=None, warn
...
+        repomd_path = Path(path) / "repodata" / "repomd.xml"
+        if not repomd_path.exists():
+            raise FileNotFoundError("No repository found at the provided path.")
+
+        repomd = Repomd(str(repomd_path))
+        metadata_files = {record.type: record for record in repomd.records}
+        parser = RepositoryReader()
+        parser._primary_xml_path = os.path.join(path, metadata_files["primary"].location_href)
+        parser._filelists_xml_path = os.path.join(path, metadata_files["filelists"].location_href)
+        parser._other_xml_path = os.path.join(path, metadata_files["other"].location_href)
```
jan-kolarik, line 420: Hmm, what about creating some structure for passing these metadata file paths? I see it on many places in the script and also in dnf the filelists are already optional, so user shouldn't be forced to fill it I guess.

dralley: I figured that `parser.repomd` (the repomd file) is that data structure already

jan-kolarik: I was rather thinking of using some single wrapper for all these *_path here, in `from_metadata_files`, in `PackageIterator`, like everywhere these paths are on the input/output. It's just an idea. But anyway, the filelists arg in `from_metadata_files` should be optional at least...

dralley: It can't be made optional currently because the underlying C API doesn't make filelists optional.  It can, eventually, but it's not currently done that way.

jan-kolarik: I am sorry, I need to look into it deeper again, but probably I'll have a chance yet in the next week.

jan-kolarik: I've created a separate issue for this: https://github.com/rpm-software-management/createrepo_c/issues/423.

examples/python/simple_repository_writing.py
```diff
@@ -0,0 +1,66 @@
...
+
+    writer.finish()
+
+
+if __name__ == "__main__":
+    if len(sys.argv) != 2 or not os.path.isdir(sys.argv[1]):
+        print("Usage: %s <directory>" % (sys.argv[0]))
+        sys.exit(1)
+
+    create_repo(sys.argv[1])
```
kontura, line 64: The `create_repo()` is not is not defined. It should probably use one of the `write_repository_v1/2`?

examples/python/simple_repository_writing.py
```diff
@@ -0,0 +1,66 @@
...
+    writer = cr.RepositoryWriter(
+        path,
+        unique_md_filenames=False,
+        changelog_limit=4,
+        checksum_type=cr.SHA512,
+        compression=cr.GZ_COMPRESSION,
+    )
+    writer.repomd.set_num_of_pkgs(len(pkg_list))
+    writer.repomd.add_repo_tag("Fedora 34")
+    writer.set_revision("1628310033")
```
kontura, line 50: `set_num_of_pkgs` is not on `repomd` but `set_revision` is.

It should probably be:
```python
    writer.set_num_of_pkgs(len(pkg_list))
    writer.repomd.add_repo_tag("Fedora 34")
    writer.repomd.set_revision("1628310033")

```

kontura (commented):
I found a couple of issues in one of the examples.

dralley:
Updated

kontura:
Thank you!

src/python/createrepo_c/__init__.py
dralley, line 674: @kontura I didn't notice this until now (sorry), but `add_repomd_metadata` is a bit redundant.  Should we fix that name?  Or is it fine because we're adding metadata to "repomd"?

src/python/createrepo_c/__init__.py
```diff
@@ -382,6 +386,364 @@ def __init__(self, primary_path, filelists_path, other_path, newpkgcb=None, warn
...
+            self.set_num_of_pkgs(num_packages)
+
+        self._updaterecords = []
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, exc_tb):
+        if exc_type:
+            pass
```
dralley, line 617: Also I think I probably should have made this raise any errors rather than swallowing them

dralley: Actually maybe they get re-raised?

src/python/createrepo_c/__init__.py
kontura: So to change it to just `add_metadata`?
I think I would like that a little bit more but I am also fine with the current form.

If you want to change it I believe we still can but the sooner the better.

src/python/createrepo_c/__init__.py
kontura: I think they do, I tried this example:
```python
class Divide:
    def __init__(self, num1, num2):
        self.num1 = num1
        self.num2 = num2

    def __enter__(self):
        return self

    def __exit__(self, exc_type, exc_value, traceback):
        pass

    def divide_by_zero(self):
        # can cause ZeroDivisionError exception
        print(self.num1 / self.num2)

# will raise a ZeroDivisionError
with Divide(3, 0) as r:
    r.divide_by_zero()
```
from https://www.geeksforgeeks.org/__exit__-in-python/ and the exception isn't swallowed.

Issue #275, opened 2021-07-13
Critical: Failed to detect compression for file package.rpm after a commit from #263 got merged; the error is returned when pkglist option is used


bkhomuts:
Hello,

After [one of recent changes](https://github.com/rpm-software-management/createrepo_c/pull/268/commits/626dbe36af8aa3540a27e364d6ca2be097b285aa), got merged, an error started appearing:

> Critical: Failed to detect compression for file package.rpm doesn't exists or not a regular file.

Where package.rpm is one of entries from the file, path for which is passed to `pkglist` option.
Createrepo_c is called with the following options:
`createrepo_c /rpm_directory --outputdir=output_repository --pkglist=package_list`

I took a look and found source of the problem. In this case, the problem is caused by a recent change of allowed_modulemd_module_metadata_file of createrepo_c.c file at https://github.com/rpm-software-management/createrepo_c/pull/268/commits/626dbe36af8aa3540a27e364d6ca2be097b285aa

The problem is that `cr_detect_compression` called with filename parameter, path to which can be relative to `directory_to_index` argument of createrepo_c. The relative path is neither converted to absolute, nor the working directory is changed to `directory_to_index` to make sure the file can be found.

Next, there is a significant performance decrease as `cr_detect_compression` is now invoked for every file. So the recommendation in this case would be to call this function only in case file name matches certain pattern, for example, `modulemd(-defaults)?.yaml(\..+)?`.

Please consider resolving the mentioned problems.

With best regards,
Bohdan Khomutskyi
Software Engineer
Red Hat

kontura:
Hello,

thank you for spotting this!
I am working on a solution, please take a look here: https://github.com/rpm-software-management/createrepo_c/pull/276

dralley:
Should this be closed now that #276 is merged?

Conan-Kudo:
Yeah.

PR #276, opened 2021-07-14
Simplify checking for allowed modulemd metadata files

kontura:
Previous approach was too complicated, slow and ultimately unnecessary.
If an invalid file passes as allowed it is caught later when loading the
module metadata. There is no need to slow down checking of each
path/package with compression validation.

For: #275

I have also included a couple of memory leaks reported by covscans, 
these usually happen in error handling so they are not a big priority.


kontura:
Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1015

Conan-Kudo (approved):

PR #279, opened 2021-08-03
Add new API for streaming parsing xml metadata

kontura:
Adds `cr_xml_parse_main_metadata_together` to both c and python APIs.
The usage is similar to `cr_xml_parse_primary`, the user can specify callbacks which get called when a full package is ready (is fully parsed).

The point is parsing all 3 metadata files at the same time and keeping only a couple of packages in memory (not the whole metadata).

At this point I think it could also handle metadata with different order of packages but the used memory would grow (we have to keep the package until its fully parsed).

@dralley can you take a look?


dralley:
@kontura Wow that was faster than expected!  We'll take a look at this this week.

examples/python/repodata_parsing.py
```diff
@@ -211,6 +211,47 @@ def newpkgcb(pkgId, name, arch):
...
+    def pkgcb(pkg):
+        # Called when whole package entry from all 3 metadata xml files is parsed
+        print_package_info(pkg)
+
+    cr.xml_parse_main_metadata_together(primary_xml_path,
+                                        filelists_xml_path,
+                                        other_xml_path,
+                                        None,
+                                        pkgcb,
+                                        warningcb)
```
dralley, line 254: The following proposal is out of scope of this PR, but I will likely make a follow-up PR that just provides an interface for all of these parsing methods.

```python
with cr.RepositoryReader("path/to/repo") as reader:
     # streaming API
     reader.for_each_package(pkgcb, ...)   

     # traditional API
     packages = reader.parse_packages(pkgcb, ...) 
```

tests/python/tests/test_xml_parser.py
```diff
@@ -950,3 +951,208 @@ def test_xml_parser_repomd_repo01_nowarningcb(self):
...
+        def pkgcb(pkg):
+            userdata["pkgcb_calls"] += 1
+
+        def warningcb(warn_type, msg):
+            userdata["warnings"].append((warn_type, msg))
+
+        cr.xml_parse_main_metadata_together(REPO_01_PRIXML, REPO_01_FILXML, REPO_01_OTHXML, newpkgcb, pkgcb, warningcb)
+
+        self.assertEqual([pkg.name for pkg in userdata["pkgs"]],
+            ['super_kernel'])
```
dralley, line 976: Formatting

tests/test_xml_parser_main_metadata_together.c
```diff
@@ -0,0 +1,162 @@
...
+
+int
+main(int argc, char *argv[])
+{
+    g_test_init(&argc, &argv, NULL);
+
+    g_test_add_func("/xml_parser_filelists/test_cr_xml_parse_main_metadata_together_00",
+                    test_cr_xml_parse_main_metadata_together_00);
+
+    //TODO(amatej): add a test where filelists/other are short and they parse faster the primary
```
dralley, line 160: Is that expected to make a difference?  Are they parsed in parallel threads?

src/python/xml_parser-py.c
```diff
@@ -73,13 +67,16 @@ c_newpkgcb(cr_Package **pkg,
 
     if (result == Py_None) {
         *pkg = NULL;
-        data->py_pkg = NULL;
-        Py_DECREF(result);
```
dralley, line 77: I understand the change on 76, less so the change on 77, is it because `Py_None`  is a singleton value and therfore the refcount doesn't need to be managed if `result` is `Py_None`?

src/xml_parser_main_metadata_together.c
```diff
@@ -0,0 +1,371 @@
...
+{
+    assert(pkg && *pkg == NULL);
+    assert(!err || *err == NULL);
+
+    cr_CbData *cb_data = cbdata;
+
+    gpointer pkey = NULL;
+    gpointer pval = NULL;
+    if (g_hash_table_lookup_extended(cb_data->in_progress_pkgs, pkgId, &pkey, &pval)) {
+        // package was already encountred in some other metadata type
```
dralley, line 100: typo "encountred"

src/xml_parser_main_metadata_together.c
```diff
@@ -0,0 +1,371 @@
...
+        } else {
+            cr_package_copy_into(pkg, pval);
+            cr_package_free(pkg);
+            pkg = pval;
+        }
+    } else {
+        // we are handling never before seen package
+
+        if (cb_data->newpkgcb) {
+            // call user specified new function if provided and copy q
```
dralley, line 168: Was this comment cut off?

src/xml_parser_main_metadata_together.c
```diff
@@ -0,0 +1,371 @@
...
+static int
+primary_pkgcb(cr_Package *pkg, void *cbdata, G_GNUC_UNUSED GError **err)
+{
+    cr_CbData *cb_data = cbdata;
+    cb_data->primary_count++;
+
+    gpointer pkey = NULL;
+    gpointer pval = NULL;
+    if (g_hash_table_lookup_extended(cb_data->in_progress_pkgs, pkg->pkgId, &pkey, &pval)) {
+        // package was already encountred in some other metadata type
```
dralley, line 153: Same typo

src/xml_parser_main_metadata_together.c
```diff
@@ -0,0 +1,371 @@
...
+        if (cb_data->newpkgcb) {
+            // call user specified new function if provided and copy q
+            cr_Package *user_created_pkg;
+            if (cb_data->newpkgcb(&user_created_pkg, pkg->pkgId, NULL, NULL, cb_data->newpkgcb_data, err)) {
+                return CR_CB_RET_ERR;
+            } else {
+                cr_package_copy_into(pkg, user_created_pkg);
+                cr_package_free(pkg);
+                pkg = user_created_pkg;
+            }
```
dralley, line 176: There's probably a good reason we need to copy the metadata instead of parsing it into the existing package - probably something to do with how primary.xml package records have a different structure from the others which don't include the pkgid as an attribute, yes?  Or something entirely unrelated?

src/xml_parser_main_metadata_together.c
```diff
@@ -0,0 +1,371 @@
...
+            // call user specified new function if provided
+            if (cb_data->newpkgcb(pkg, pkgId, name, arch, cb_data->newpkgcb_data, err)) {
+                return CR_CB_RET_ERR;
+            }
+        } else {
+            *pkg = cr_package_new();
+        }
+        g_hash_table_insert(cb_data->in_progress_pkgs, g_strdup(pkgId), *pkg);
+    }
+
```
dralley, line 122: I would like to better understand the distinction between this `newpkgcb_general` and `primary_pkgcb`.  As far as I can tell, one distinction is that `newpkgcb_general` is used for filelists and other but *not* primary, because primary doesn't provide pkgid as an attribute - so handling that leads to some code duplication.

Is that correct or is it a poor / incomplete understanding?

Would it be valid for this callback to set `pkg->loadingflags |= CR_PACKAGE_FROM_XML;` here?

src/xml_parser_main_metadata_together.c
```diff
@@ -0,0 +1,371 @@
...
+    pkg->loadingflags |= CR_PACKAGE_FROM_XML;
+    pkg->loadingflags |= CR_PACKAGE_LOADED_PRI;
+
+    return call_user_callback_if_package_finished(pkg, cb_data, err);
+}
+
+static int
+parse_next_section(CR_FILE *target_file, const char *path, cr_ParserData *pd, GError **err)
+{
+    char buf[XML_BUFFER_SIZE];
```
dralley, line 190: So, instead of parsing from start to end element of a package one package at a time, we parse the file one chunk of bytes at a time, which could be >1 package or <1 package, and just keep track of which packages are "done" -- yes?

src/xml_parser_main_metadata_together.c
```diff
@@ -0,0 +1,371 @@
...
+    //TODO(amatej): if newpkgcb return NULL we should skip the pkg...
+    primary_pd = new_primary_parser_data(newpkgcb_primary, &cbdata, primary_pkgcb, &cbdata, warningcb, warningcb_data, 0);
+    cr_ParserData *filelists_pd;
+    filelists_pd = new_filelists_parser_data(newpkgcb_general, &cbdata, filelists_pkgcb, &cbdata, warningcb, warningcb_data);
+    cr_ParserData *other_pd;
+    other_pd = new_other_parser_data(newpkgcb_general, &cbdata, other_pkgcb, &cbdata, warningcb, warningcb_data);
+
+    char primary_is_done = 0;
+    char filelists_is_done = 0;
+    char other_is_done = 0;
```
dralley, line 278: `char` capacity is 127, `XML_BUFFER_SIZE` is `8192`, which means the average package element size in the buffer in any of the 3 files *must* be >64 bytes.  This is a safe assumption but if `XML_BUFFER_SIZE` ever increased in the future (by a factor of 3 or 4) then it might result in an overflow.  Maybe uint32 should be used just to be safe.

src/xml_parser_main_metadata_together.c
```diff
@@ -0,0 +1,371 @@
...
+    while (1) {
+        if (!primary_is_done) {
+            do {
+                primary_is_done = parse_next_section(primary_f, primary_path, primary_pd, err);
+                if (*err) {
+                    ret = (*err)->code;
+                    goto out;
+                }
+
+            //TODO(amatej): what about not providing filelists? well at least one of them should always be provided
```
dralley, line 291: Unrelated question.  What is the use case for having a subset of the files listed in primary.xml?  Is the subset enough for reliable depsolving?  What tools (if any) use that information and for what purposes?

And how was the heuristic decided?  https://github.com/rpm-software-management/createrepo_c/blob/8eff6ed99f5fd0ba844cb8513963435caab5fd3c/src/misc.h#L105-L118

src/xml_parser_main_metadata_together.c
```diff
@@ -0,0 +1,371 @@
...
+            g_propagate_error(err, filelists_pd->err);
+            break;
+        }
+        if (other_pd->err) {
+            ret = other_pd->err->code;
+            g_propagate_error(err, other_pd->err);
+            break;
+        }
+
+        //TODO(amatej): add tests for out of order metadata
```
dralley, line 331: For what it's worth, as mentioned in the email we're not aware of any tools that can generate such metadata, and Pulp 2 would not have been capable of parsing it at all.  We've not had any reports of that issue, so such out-of-order metadata may not exist in the wild.

src/xml_parser_main_metadata_together.c
```diff
@@ -0,0 +1,371 @@
...
+        //TODO(amatej): fix memory leaks in error handling (tests)
+        if (primary_is_done && filelists_is_done && other_is_done)
+            break;
+        if (ret != CRE_OK)
+            break;
+    }
+
+out:
+    if (ret != CRE_OK) {
+        // An error already encoutentered
```
dralley, line 341: Typo "encoutentered"

src/xml_parser_main_metadata_together.c
```diff
@@ -0,0 +1,371 @@
...
+                    path,
+                    (xml_err) ? (int) xml_err->line : 0,
+                    (xml_err) ? (char *) xml_err->message : "UNKNOWN_ERROR");
+        return -1;
+    }
+
+    return done;
+}
+
+//TODO(amatej): add a flag that enforces pkg are in the same order (If possible?)
```
dralley, line 214: For what it's worth, as mentioned in the email we're not aware of any tools that can generate such metadata, and Pulp 2 would not have been capable of parsing it at all. We've not had any reports of that issue, so such out-of-order metadata may not exist in the wild.  To be honest we might even consider turning on this flag.

I feel like the only obvious way it could happen is a "torn" repository with simple metadata filenames.

src/xml_parser_main_metadata_together.c
dralley: This makes a lot of sense given the design constraints you're dealing with (libxml, existing parsers written in SAX style).  I wasn't sure whether it could be done without a "pull" / "StAX" parser, which I don't think libxml provides, so I thought it might require a more fundamental change to the architecture.  

There's some limitations to the approach, but it will work, and it's nice to see that it didn't end up requiring a complete rewrite.

src/xml_parser_main_metadata_together.c
```diff
@@ -0,0 +1,371 @@
...
+}
+
+static int
+filelists_pkgcb(cr_Package *pkg, void *cbdata, G_GNUC_UNUSED GError **err)
+{
+    cr_CbData *cb_data = cbdata;
+    cb_data->filelists_count++;
+    pkg->loadingflags |= CR_PACKAGE_LOADED_FIL;
+    return call_user_callback_if_package_finished(pkg, cb_data, err);
+}
```
dralley, line 133: Same kind of question here.  I assume that for the same reason, the package callback has to copy data into the package while the filelist and other package callbacks do not?

src/python/xml_parser-py.c
dralley: Never mind - I think it's because PyDict_SetItem increments the count of `result`, but we're done with it here, so (new) line 78 handles both codepaths.

src/python/xml_parser-py.c
```diff
@@ -91,16 +88,24 @@ c_pkgcb(cr_Package *pkg,
...
+        result = PyObject_CallObject(data->py_pkgcb, arglist);
+        Py_DECREF(arglist);
+        PyDict_DelItem(data->py_pkgs, keyFromPtr);
+    } else {
+        // The package was not provided by user in c_newpkgcb,
+        // create new python package object
+        PyObject *new_py_pkg = Object_FromPackage(pkg, 1);
+        arglist = Py_BuildValue("(O)", new_py_pkg);
+        result = PyObject_CallObject(data->py_pkgcb, arglist);
+        Py_DECREF(arglist);
```
dralley, line 104: `arglist` could be decref'd once at the end rather than on both branches.

examples/python/repodata_parsing.py
dralley: Which... is pretty close to what the first "first_method" API does.  Why is using it discouraged?

dralley:
@kontura I'm done reviewing ^^

dralley:
Finally got around to actually measuring the difference.  When parsing the RHEL6 metadata, the old strategy peaked at 1.9gb memory, while the new one peaks at 19 megabytes.  Excellent!

Old

Screenshot from 2021-08-30 00-49-59

New

Screenshot from 2021-08-30 00-40-14

dralley:
@kontura I've created a PR updating Pulp to use this API and the tests are passing.  I haven't seen any issues thus far.

Let me know if there's anything else I can do to help, we would love to start using this soon.

examples/python/repodata_parsing.py
kontura: I don't know.

Maybe because it can only read all the repodata into memory whereas the `xml_parser_primary` for example can avoid that (if you need just the primary, so not in this case)?

src/xml_parser_main_metadata_together.c
kontura: You are correct. To be honest I didn't know that "pull" / "StAX" parsing models were a thing. Now I managed to find this libxml2 API: http://xmlsoft.org/xmlreader.html, I had no idea about this. :slightly_frowning_face: 
It would be a big rewrite but it looks like its the better solution and it seems DTD validation is also easier with that approach.

I will look into it some more, but I am not sure if I will be able to do it right now.

src/xml_parser_main_metadata_together.c
kontura: Good questions I will look into this more.
Maybe @dmach could know?

I think its not enough for all depsolving, there are use cases where dnf needs to download the filelists.

src/xml_parser_main_metadata_together.c
kontura: I was also thinking it might be useful.

src/xml_parser_main_metadata_together.c
kontura: Yep, that is right.

src/python/xml_parser-py.c
kontura: I will move the `Py_DECREF` after that branch, thanks. :slightly_smiling_face:

kontura:
Sorry I was offline for a some time. Nice it looks good.

I also tried to do some basic speed comparing and it seems pretty much the same as the old all in memory approach.

src/xml_parser_main_metadata_together.c
dralley: No worries, I agree that it would basically represent a complete rewrite.  This is probably the better  solution in the short term and there's now that we have it.

I was actually experimenting with this method in a side project written in Rust, but I haven't really had time to work on it in months, so it's still pretty limited (no streaming one package at a time, although it would definitely be possible eventually).  I think it would mostly translate to C (if you eventually wanted to use it as a reference).  

https://github.com/dralley/rpmrepo_rs/blob/master/rpmrepo_metadata/src/primary.rs#L153-L198
https://github.com/dralley/rpmrepo_rs/blob/master/rpmrepo_metadata/src/primary.rs#L400-L604

src/xml_parser_main_metadata_together.c
dralley: I misunderstood what they were for, then

dralley:
re: speed, I reached the same conclusion.

dralley:
@kontura Do you have an approximate timeline for when we can expect this to be ready?  We're under a little pressure to reduce memory consumption further.

kontura:
@dralley I see, I would like to finish this in the next two weeks or so.

src/xml_parser_main_metadata_together.c
```diff
@@ -0,0 +1,371 @@
...
+static int
+call_user_callback_if_package_finished(cr_Package *pkg, cr_CbData *cb_data, GError **err)
+{
+    if ((pkg->loadingflags & CR_PACKAGE_LOADED_PRI) &&
+        (pkg->loadingflags & CR_PACKAGE_LOADED_OTH) &&
+        (pkg->loadingflags & CR_PACKAGE_LOADED_FIL))
+    {
+        g_hash_table_remove(cb_data->in_progress_pkgs, pkg->pkgId);
+
+        // call user packace callback
```
dralley, line 51: Typo `packace`

src/package_internal.h
```diff
@@ -0,0 +1,39 @@
+/* createrepo_c - Library of routines for manipulation with repodata
+ * Copyright (C) 2013  Tomas Mlcoch
```
Conan-Kudo, line 2: I don't think you're @Tojaj...?

Conan-Kudo (changes requested):

src/package_internal.h
kontura: sorry, it should be fixed now.

Conan-Kudo (approved):

examples/python/repodata_parsing.py
dralley: @dmach Do you know?

src/xml_parser_main_metadata_together.c
dralley: @dmach I know you've moved on, but do you happen to know the answer to this question?  Or who I could ask?

j-mracek: I suggest that the code is related to the future requirement to not loading/downloading filelists. When some package requires a file it is mostly located in ets, bin, user/bin, therefore to add all files from this location will allow to not require filelist in future (in theory).

src/xml_parser_main_metadata_together.c
dralley: FYI I implemented the Rust version of package streaming one-by-one in a couple of hours.

PR #280, opened 2021-08-16
Track if merger is empty and don't try to dump index if it is

kontura:
Libmodulemd errors out when dumping empty index. This can happen if
running with just `--update` (no `--keep-all-metadata`) and old repo
has modules metadata which habe been removed in the meantime.

Test: https://github.com/rpm-software-management/ci-dnf-stack/pull/1027

### DIFF
--- a/src/createrepo_c.c
+++ b/src/createrepo_c.c
@@ -816,6 +816,7 @@ main(int argc, char **argv)
 #ifdef WITH_LIBMODULEMD
     // module metadata found in repo
     if (cmd_options->modulemd_metadata) {
+        gboolean merger_is_empty = TRUE;
         ModulemdModuleIndexMerger *merger = modulemd_module_index_merger_new();
         if (!merger) {
             g_critical("Could not allocate module merger");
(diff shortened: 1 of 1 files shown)


src/createrepo_c.c
```diff
@@ -880,48 +882,53 @@ main(int argc, char **argv)
...
+            new_modules_metadatum->type = g_strdup("modules");
+            additional_metadata = g_slist_prepend(additional_metadata, new_modules_metadatum);
         }
 
-        //create additional metadatum for new module metadata file
-        cr_Metadatum *new_modules_metadatum = g_malloc0(sizeof(cr_Metadatum));
-        new_modules_metadatum->name = modules_metadata_path;
-        new_modules_metadatum->type = g_strdup("modules");
-        additional_metadata = g_slist_prepend(additional_metadata, new_modules_metadatum);
+        g_clear_pointer(&merger, g_object_unref);
```
pkratoch, line 930: Why did you move clearing the merger to the end? This way, the code can exit without clearing it first in case of failure. Or am I missing something?

pkratoch:
Also, there is a typo in the commit message: habe -> have
Have you switched to German for a second there? :)

src/createrepo_c.c
pkratoch: O right, it must be cleared in case merger_is_empty. But it still holds that the pointer is not cleared in case it exits earlier during the if.

kontura: Fixed :+1:

kontura:
> Also, there is a typo in the commit message: habe -> have
> Have you switched to German for a second there? :)

Changed it back to English :smile:, thanks.

PR #284, opened 2021-09-09
Report the exact error when loading of metadata fails.

velichkov:


src/mergerepo_c.c
```diff
@@ -710,9 +710,10 @@ merge_repos(GHashTable *merged,
 
         g_debug("Processing: %s", repopath);
 
-        if (cr_metadata_load_xml(metadata, ml, NULL) != CRE_OK) {
+        if (cr_metadata_load_xml(metadata, ml, &err) != CRE_OK) {
```
kontura, line 713: Right now this will cause a problem when doing a build without libmodulemd because `err` is defined only when `WITH_LIBMODULEMD` is used.
I think we should move the`err` definition out of the `ifdef WITH_LIBMODULEMD` scope.

kontura (changes requested):
I have found one problem please take a look.
Otherwise this looks good to me. :+1:

src/mergerepo_c.c
velichkov: Done

kontura (approved):
Great, thank you!

PR #287, opened 2021-09-23
Default --keep-all-metadata to TRUE and add --discard-additional-metadata

kontura:
= changelog =
msg: Switch default of --keep-all-metadata to TRUE and add --discard-additional-metadata
type: enhancement
resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1992209

`--keep-all-metadata` controls whether to keep additional metadata (such as comps, updateinfo, modules..) during a run with `--update`. With this PR they will be kept by default whereas before `--keep-all-metadata` had to be used.

We might want this because modular repositories really break without the `modules.yaml`.

### DIFF
--- a/doc/createrepo_c.8
+++ b/doc/createrepo_c.8
@@ -1,6 +1,6 @@
 .\" Man page generated from reStructuredText.
 .
-.TH CREATEREPO_C 8 "2020-07-02" "" ""
+.TH CREATEREPO_C 8 "2021-09-23" "" ""
 .SH NAME
 createrepo_c \- Create rpm-md format (xml-rpm-metadata) repository
 .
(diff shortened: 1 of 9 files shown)


Conan-Kudo:
I think this makes sense. We already do this implicitly with modifyrepo and mergerepo commands, so I don't see why we wouldn't do it for createrepo too.

Conan-Kudo:
@kontura have you considered incorporating #73 into this, though? It'd be a nice improvement...

kontura:
> @kontura have you considered incorporating #73 into this, though? It'd be a nice improvement...

I have not, but looking at it now I think that the chosen approach is not quite right. https://github.com/rpm-software-management/createrepo_c/pull/73#issuecomment-284488761 describes a lot of issues.

However maybe instead of checking whether each package was used from cache we could be more precise by comparing the old and the new repomd [structs](https://github.com/rpm-software-management/createrepo_c/blob/master/src/repomd.h#L103-L117). We could check that all the repomd records are identical (their hashes match) as well as all the other `repo_tags`, `content_tags`...

@Conan-Kudo what do you think? I will try it out.

Conan-Kudo:
That makes sense to me, @kontura!

kontura:
I have added a few commits with the discussed functionality.
The tests were also updated.

Conan-Kudo:
@kontura It looks like a few integration tests are failing. Are they expected to fail?

kontura:
They are, it is mostly the change of default.
I have a CI update PR as well: https://github.com/rpm-software-management/ci-dnf-stack/pull/1037

Conan-Kudo:
I merged the tests PR and I've re-run the CI jobs. We'll see how that goes.

Conan-Kudo (approved):

Issue #286, opened 2021-09-20
createrepo_c --update can corrupt dependencies


mlschroe:
The centos-8 bpg-algeti-fonts-2.005-11.el8.noarch.rpm package contains the provides `font(bpgalgetigpl&gnu)`.
This ends up in primary.xml file like this:
```
<rpm:entry name="font(bpgalgetigpl&amp;gnu)"/>
```

For some unknown reason, libxml2 re-escapes '&' characters in attributes (unlike libexpat). Any `&` character will be replaced by `&#38;`. This means that a createrepo_c --update operation will write a bad dependency line:
```
<rpm:entry name="font(bpgalgetigpl&amp;#38;gnu)"/>
```

(libsolv suffers from the same issue. I fixed it with https://github.com/openSUSE/libsolv/commit/9a13593b43bd883be6bf4260edbfc8eaf766ffa8)

kontura:
I have made a PR: https://github.com/rpm-software-management/createrepo_c/pull/289
It uses a different approach than libsolv.

dralley:
Ultimately the issue is not that it re-encodes the entities but that they are not decoded by the parsing in the first place.  We're reading these values and then saving them to the database, and they're incorrect (because the encoding wasn't reversed)

PR #289, opened 2021-11-08
Enable entity substitution when parsing xmls

kontura:
Repodata don't use any custom entities but libxml2 enforces some in
order to have a valid xml document when saved (`<`, `>`, `&`,...)
http://xmlsoft.org/entities.html

When createrepo_c runs with `--update` it parses repodata which can
contain such entities (`&amp;` for `&`). We want to substitute them back
so when the data is used again it doesn't get re-escaped.

https://github.com/rpm-software-management/createrepo_c/issues/286

### DIFF
--- a/src/xml_parser.c
+++ b/src/xml_parser.c
@@ -292,3 +292,47 @@ cr_xml_parser_generic_from_string(xmlParserCtxtPtr parser,
 
     return ret;
 }
+
+const xmlChar **
+unescape_ampersand_from_values(const xmlChar **attr, gboolean *allocation_needed) {
+    *allocation_needed = FALSE;
+
+    if (!attr) {
+        return attr;
+    }
+
+    // In the vast majority of cases there is no '&' in the
+    // data so check if there is any first and if not return
+    // quickly.
+    //
+    // attr consists of key1, value1, key2, value2 pairs,
+    // we know which keys we want (they don't contain &) so
+    // we don't have to check those.
+    size_t nattr;
+    for (nattr = 1; attr[nattr]; nattr+=2) {
+        if (strchr((char *)attr[nattr], '&')) {
+            *allocation_needed = TRUE;
+        }
+    }
+
+    if (!*allocation_needed) {
+        return attr;
+    }
+
+    char **attr_copy = g_malloc0(sizeof(char *) * (nattr - 1));
+    if (attr_copy) {
+        for (nattr = 0; attr[nattr]; nattr++) {
+            if (strchr((char *)attr[nattr], '&')) {
+                char **cut_out_amp = g_strsplit((char *)attr[nattr], "#38;", -1);
+                attr_copy[nattr] = g_strjoinv(NULL, cut_out_amp);
+                g_strfreev(cut_out_amp);
+            } else {
+                attr_copy[nattr] = g_strdup((char *)attr[nattr]);
+            }
+        }
+
+        attr_copy[nattr] = 0;
+    }
+
+    return (const xmlChar **)attr_copy;
+}
--- a/src/xml_parser_filelists.c
+++ b/src/xml_parser_filelists.c
@@ -90,6 +90,9 @@ cr_start_handler(void *pdata, const xmlChar *element, const xmlChar **attr)
         return;
     }
 
+    gboolean free_attr = FALSE;
+    attr = unescape_ampersand_from_values(attr, &free_attr);
+
     // Update parser data
     pd->state      = sw->to;
     pd->docontent  = sw->docontent;
@@ -196,6 +199,10 @@ cr_start_handler(void *pdata, const xmlChar *element, const xmlChar **attr)
 
     default:
         break;
+    }
+
+    if (free_attr) {
+        g_strfreev((char **)attr);
     }
 }
 
--- a/src/xml_parser_internal.h
+++ b/src/xml_parser_internal.h
@@ -249,6 +249,15 @@ other_parser_data_new(cr_XmlParserNewPkgCb newpkgcb,
                       cr_XmlParserWarningCb warningcb,
                       void *warningcb_data);
 
+/** Replace &#38; by real ampersand char from values in attr.
+ * @param attr                   List of attributes
+ * @param allocation_needed      Output bool whether returned attr has to be freed.
+ * @return                       attr with regular '&' instead of "&#38";
+ */
+const xmlChar **unescape_ampersand_from_values(const xmlChar **attr,
+                                               gboolean *allocation_needed);
+
+
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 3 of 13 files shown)


kontura:
Test: https://github.com/rpm-software-management/ci-dnf-stack/pull/1053

mlschroe:
You shouldn't do it that way AFAIK as XML_PARSE_NOENT is problematic security wise. You mustn't allow libxml to load external entities.

mlschroe:
See for example https://access.redhat.com/security/cve/cve-2014-0179

kontura:
> You shouldn't do it that way AFAIK as XML_PARSE_NOENT is problematic security wise. You mustn't allow libxml to load external entities.

I see, that is good to know, thanks.

I spent some more time digging through libxml2 but it seems there is no other way to avoid the `&#38;`.
I have updated the PR with the same approach as is used in libsolv.

dralley:
I went looking through the libxml documentation and, sure enough, there does not seem to be any general-purpose function for decoding basic xml encoded strings.  That is... kind of a mind-blowing omission, considering it does have the counterpart functionality.

http://xmlsoft.org/html/libxml-entities.html#xmlEncodeSpecialChars

src/xml_parser.c
```diff
@@ -292,3 +292,47 @@ cr_xml_parser_generic_from_string(xmlParserCtxtPtr parser,
 
     return ret;
 }
+
+const xmlChar **
+unescape_ampersand_from_values(const xmlChar **attr, gboolean *allocation_needed) {
```
dralley, line 297: It sounds like there's a hard example for `&`, and addressing that is probably enough for now, but is it theoretically possible for other special characters to sneak in?

kontura: I am not sure what is allowed, [here](https://rpm-software-management.github.io/rpm/manual/spec.html) it says: "Names should not include any numeric operators (‘<’, ‘>’,’=’)" but that's names not capabilities. For capabilities there is just: "Capability names must start with alphanumerics or underscore".

However I tried it with `<` and those don't get encoded like `&` in libxml2. It seems `&` is a special case: https://github.com/GNOME/libxml2/blob/master/parser.c#L3954

dralley:
@kontura We do need this fixed, the data gets saved to our database, so it's difficult to clean up.  I didn't realize we were impacted but we are.

kontura:
> @kontura We do need this fixed, the data gets saved to our database, so it's difficult to clean up. I didn't realize we were impacted but we are.

I think this PR should be ready.
Not sure about the labels. What are the requested changes @Conan-Kudo?

dralley:
A test for this would be great, as well.  That might require some new fixture data which tries to put the encoded entities in a bunch of different places in the metadata.

One random unrelated question I have: do you know if the XML spec mandates that tab characters `\t` in attribute names are substituted with spaces?  I noticed that createrepo_c does this but I can't figure out where, or why.  I looked through the libxml2 source code and the XML specs and couldn't find any reference to this.

kontura:
> A test for this would be great, as well. That might require some new fixture data which tries to put the encoded entities in a bunch of different places in the metadata.

We already have a [PR](https://github.com/rpm-software-management/ci-dnf-stack/pull/1053) for a CI test but I did create a couple of unittests for just the xml parser and metadata sprinkled with `&amp;` Added in  a new commit.


> One random unrelated question I have: do you know if the XML spec mandates that tab characters `\t` in attribute names are substituted with spaces? I noticed that createrepo_c does this but I can't figure out where, or why. I looked through the libxml2 source code and the XML specs and couldn't find any reference to this.

I don't know but I am also not sure I understand. Do you mean in the attribute values? something like:
```
...
<collection short="foo.com\tponent">
...
```
or do you mean in the `sho\trt`? Can you provide an example?

Conan-Kudo:
Do we have local entity files to make this work? It sounds like we need to start having them as part of createrepo_c...

dralley:
@kontura example from the changelogs:

`('Jeff Johnston\t<jjohnstn@redhat.com> 1:6.0.1-7', 1260532801, '- Resolves: #546784\n- Rebase Autotools to Linux tools 0.4.0 release plus patches.\n- Rebase Libhover to Linux tools 0.4.0 release.\n- Remove Libhover patch which is already part of rebase.')`

`('Jeff Johnston <jjohnstn@redhat.com> 1:6.0.1-7', 1260532801, '- Resolves: #546784\n- Rebase Autotools to Linux tools 0.4.0 release plus patches.\n- Rebase Libhover to Linux tools 0.4.0 release.\n- Remove Libhover patch which is already part of rebase.')`

kontura:
> @kontura example from the changelogs of centos7:
> ...


> @kontura example from the changelogs of centos7:
> 
> original values as per XML (except \t is an actual tab character) --------------------------v
> 
> `('Jeff Johnston\t<jjohnstn@redhat.com> 1:6.0.1-7', 1260532801, '- Resolves: #546784\n- Rebase Autotools to Linux tools 0.4.0 release plus patches.\n- Rebase Libhover to Linux tools 0.4.0 release.\n- Remove Libhover patch which is already part of rebase.')`
> 
> after parsing
> 
> `('Jeff Johnston <jjohnstn@redhat.com> 1:6.0.1-7', 1260532801, '- Resolves: #546784\n- Rebase Autotools to Linux tools 0.4.0 release plus patches.\n- Rebase Libhover to Linux tools 0.4.0 release.\n- Remove Libhover patch which is already part of rebase.')`

I see, yea I don't know about the spec or what is required but I looked and createrepo_c is not doing this it happens in libxml2. I think it might be: https://gitlab.gnome.org/GNOME/libxml2/-/blob/master/parser.c#L4075-4078 since `0x9` is the `\t` and it is replaced by `0x20` a space. There appears to be some normalization going on.
I didn't really investigate deeper though.

kontura:
> Do we have local entity files to make this work? It sounds like we need to start having them as part of createrepo_c...

I am not sure I understand, what would we need them for?
This PR is working directly on strings and the entity substitution is turned off.

dralley:
>I see, yea I don't know about the spec or what is required but I looked and createrepo_c is not doing this it happens in libxml2. I think it might be: https://gitlab.gnome.org/GNOME/libxml2/-/blob/master/parser.c#L4075-4078 since 0x9 is the \t and it is replaced by 0x20 a space. There appears to be some normalization going on. I didn't really investigate deeper though.

Yeah that's fine it was just a side comment.  I do see now where it's happening and also, where it's described in the spec.  So this is normal and correct.

Conan-Kudo:
> > Do we have local entity files to make this work? It sounds like we need to start having them as part of createrepo_c...
> 
> I am not sure I understand, what would we need them for? This PR is working directly on strings and the entity substitution is turned off.

Okay then, I guess I'm just confused. :)

dralley:
>I think this PR should be ready.

@kontura FYI this PR is still marked as work-in-progress

dralley:
I don't feel as confident reviewing the C portion but the Python portion looks good to me.  Would it be possible to cut a release immediately after this merges (along with the removal of the old API, and the already merged patch)

Conan-Kudo (approved):

PR #292, opened 2021-11-15
Zero init buffer to prevent use of garbage values if input is too short

kontura:
Discovered by covscans.


kontura:
Added an additional commit to fix a couple more covscan warnings.

Conan-Kudo (approved):

PR #295, opened 2021-11-24
New pull-based streaming parsing api

dralley:


dralley:
Making some progress.  

`test_xml_parser.py::TestCaseXmlParserMainMetadataTogether::test_xml_parser_main_metadata_together_error` still has a leak and `test_xml_parser.py::TestCaseXmlParserMainMetadataTogether::test_xml_parser_main_metadata_together_pkgcb_abort` has a use-after-free.

Will continue on Monday.

src/xml_parser_main_metadata_together.c
```diff
@@ -336,133 +328,215 @@ int cr_xml_parse_main_metadata_together(const char *primary_path,
...
+    cr_CbData *cbdata = iter->cbdata;  // avoid dereferencing void *
+    if (iter->allow_out_of_order) {
+        g_hash_table_destroy(cbdata->in_progress_pkgs_hash);
     } else {
-        cr_slist_free_full(cbdata.in_progress_pkgs_list, (GDestroyNotify) cr_package_free);
+        cr_slist_free_full(cbdata->in_progress_pkgs_list, (GDestroyNotify) cr_package_free);
     }
 
-    return ret;
+    g_queue_free_full(cbdata->finished_pkgs_queue, (GDestroyNotify) cr_package_free);
```
dralley, line 538: The current problems - I think - are caused by a pseudo joint-ownership of the underlying package objects between `py_pkgs` and `finished_pkgs_queue`.  

The outer Python newpkgcb creates the package object and puts it into `py_pkgs`, and simultaneously they end up on `finished_pkgs_queue`.  When an error is hit during parsing `cr_parse_main_metadata_together()` cleans up the package iterator including `finished_pkgs_queue`, but then the Python wrapper tries to clean up `py_pkgs`, which deallocates all of the packages within, except the underlying cr_package has already been freed.

But I'm not 100% sure because at this point my brain is fried. Any ideas about how to resolve that one?

src/xml_parser.h
```diff
@@ -317,6 +317,43 @@ cr_xml_parse_main_metadata_together(const char *primary_path,
...
+
+    gboolean allow_out_of_order;
+    GError *tmp_err;
+
+    gboolean primary_is_done;
+    gboolean filelists_is_done;
+    gboolean other_is_done;
+
+    void *cbdata;
+} cr_PkgIterator;
```
kontura, line 341: I think it should be possible to make the content private.

Here it would be just:  `typedef struct _cr_PkgIterator cr_PkgIterator;`

And at the top xml_parser_main_metadata_together.c:
```c
struct _cr_PkgIterator {
    CR_FILE *primary_f;
    CR_FILE *filelists_f;
    CR_FILE *other_f;

    const char *primary_path;
    const char *filelists_path;
    const char *other_path;

    void *primary_pd;
    void *filelists_pd;
    void *other_pd;

    gboolean allow_out_of_order;
    GError *tmp_err;

    gboolean primary_is_done;
    gboolean filelists_is_done;
    gboolean other_is_done;

    void *cbdata;
};
```
(no `typedef` here)

AFAIK this should work, it is the same as for example `_cr_ChecksumCtx`.

src/xml_parser_main_metadata_together.c
```diff
@@ -337,133 +328,217 @@ int cr_xml_parse_main_metadata_together(const char *primary_path,
...
+        // call user package callback
+        if (pkgcb && pkgcb(package, pkgcb_data, &tmp_err)) {
+            if (tmp_err) {
+                g_propagate_prefixed_error(err, tmp_err, "Parsing interrupted: ");
+            } else {
+                g_set_error(err, ERR_DOMAIN, CRE_CBINTERRUPTED, "Parsing interrupted");
+            }
+            // cr_package_free(package)
+            cr_PkgIterator_free(pkg_iterator, err);
+            return CR_CB_RET_ERR;
```
kontura, line 353: Should we return the error code here? `return (*err)->code`.
This returns from `cr_xml_parse_main_metadata_together()` not from a callback.

src/xml_parser_main_metadata_together.c
```diff
@@ -337,133 +328,217 @@ int cr_xml_parse_main_metadata_together(const char *primary_path,
...
+        } else {
+            // If callback return CRE_OK but it simultaneously set
+            // the tmp_err then it's a programming error.
+            assert(tmp_err == NULL);
+        };
+        // cr_package_free(package)
+    }
+    assert(cr_PkgIterator_is_finished(pkg_iterator));
```
kontura, line 361: When there was an error during parsing we have the `err` variable set and parsing is not finished. We should check `err` before doing the assert.

src/xml_parser_main_metadata_together.c
```diff
@@ -337,133 +328,217 @@ int cr_xml_parse_main_metadata_together(const char *primary_path,
...
+    if (allow_out_of_order) {
+        cbdata->in_progress_pkgs_hash = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, NULL);
+    }
+
+    GError* tmp_err = new_iter->tmp_err;
+    new_iter->primary_f = cr_open(primary_path, CR_CW_MODE_READ, CR_CW_AUTO_DETECT_COMPRESSION, &tmp_err);
     if (tmp_err) {
-        ret = tmp_err->code;
         g_propagate_prefixed_error(err, tmp_err, "Cannot open %s: ", primary_path);
         g_clear_error(&tmp_err);
```
kontura, line 435: This is my mistake but the `g_clear_error(&tmp_err);` shouldn't be here (for all 3 cases of error handling).
The `g_propagate_prefixed_error()` [moves the error](https://docs.gtk.org/glib/func.propagate_prefixed_error.html) so if we later clean it it frees `err` content.

src/xml_parser_main_metadata_together.c
```diff
@@ -337,133 +328,217 @@ int cr_xml_parse_main_metadata_together(const char *primary_path,
...
-        goto out;
+        cr_PkgIterator_free(new_iter, err);
```
kontura, line 436: Passing `err` probably doesn't make sense we are already handling an error and `err` is already set.
Maybe just `cr_PkgIterator_free(new_iter, NULL);`?

kontura (commented):
Uff, all these callbacks are quite a headache. I hope it will simplify a bit when we switch to the  "pull" / "StAX" parsing model.

---

I have also noticed one more memory leak that is already present in the original code.
In `queue_package_if_finished()` there is:
```c
//remove first element in the list
cb_data->in_progress_pkgs_list = cb_data->in_progress_pkgs_list->next;
```
which I now realize is leaking the node, it probably should be something like:
```c
//remove first element in the list
cb_data->in_progress_pkgs_list = g_slist_delete_link(cb_data->in_progress_pkgs_list, cb_data->in_progress_pkgs_list);
```

dralley:
> I hope it will simplify a bit when we switch to the "pull" / "StAX" parsing model.

Is this a plan now?  :D

kontura:
> Is this a plan now? :)

I would definitely like to do it, hopefully in not so distant future. :slightly_smiling_face:

src/xml_parser.h
dralley: I actually tried this already but I get a bunch of errors like this:

```
  /home/vagrant/devel/community/createrepo_c/src/xml_parser_main_metadata_together.c:530:13: error: invalid use of incomplete typedef ‘cr_PkgIterator’ {aka ‘struct _cr_PkgIterator’}                                                       
      530 |     if (iter->tmp_err) {                                                                                                                                                                                                          
          |             ^~                                                                                                                                                                                                                    
    /home/vagrant/devel/community/createrepo_c/src/xml_parser_main_metadata_together.c:533:22: error: invalid use of incomplete typedef ‘cr_PkgIterator’ {aka ‘struct _cr_PkgIterator’}                                                       
      533 |         cr_close(iter->primary_f, NULL);                                                                                                                                                                                          
          |                      ^~                                                                                                                                                                                                           
    /home/vagrant/devel/community/createrepo_c/src/xml_parser_main_metadata_together.c:534:22: error: invalid use of incomplete typedef ‘cr_PkgIterator’ {aka ‘struct _cr_PkgIterator’}                                                       
      534 |         cr_close(iter->filelists_f, NULL);                                                                                                                                                                                        
          |                      ^~                                                                                                                                                                                                           
    /home/vagrant/devel/community/createrepo_c/src/xml_parser_main_metadata_together.c:535:22: error: invalid use of incomplete typedef ‘cr_PkgIterator’ {aka ‘struct _cr_PkgIterator’}                                                       
      535 |         cr_close(iter->other_f, NULL);                                                                                                                                                                                            
          |                      ^~                                                                                                                                                                                                           
    /home/vagrant/devel/community/createrepo_c/src/xml_parser_main_metadata_together.c:537:31: error: invalid use of incomplete typedef ‘cr_PkgIterator’ {aka ‘struct _cr_PkgIterator’}                                                       
      537 |         GError* tmp_err = iter->tmp_err;                                                                                                                                                                                          
          |                               ^~
    /home/vagrant/devel/community/createrepo_c/src/xml_parser_main_metadata_together.c:539:22: error: invalid use of incomplete typedef ‘cr_PkgIterator’ {aka ‘struct _cr_PkgIterator’}
      539 |         cr_close(iter->primary_f, &tmp_err);
          |                      ^~                                                                                    
    /home/vagrant/devel/community/createrepo_c/src/xml_parser_main_metadata_together.c:541:26: error: invalid use of incomplete typedef ‘cr_PkgIterator’ {aka ‘struct _cr_PkgIterator’}
      541 |             cr_close(iter->filelists_f, &tmp_err);
          |                          ^~
    /home/vagrant/devel/community/createrepo_c/src/xml_parser_main_metadata_together.c:544:26: error: invalid use of incomplete typedef ‘cr_PkgIterator’ {aka ‘struct _cr_PkgIterator’}
      544 |             cr_close(iter->other_f, &tmp_err);
          |                          ^~
```

kontura: I just tried it and it seems to work for me.

Looking at the errors you mentioned make sure you are not `typedef`ing the `_cr_PkgIterator` in xml_parser_main_metadata_together.c.

It is working for me if I use the exact form I have in my comment: `struct _cr_PkgIterator {.. };` not `typedef struct {...} _cr_PkgIterator;`. So it cannot be just copied from xml_parser.h.

dralley: >It is working for me if I use the exact form I have in my comment 

Yup, that's the issue, even without using `typedef` it seems like if you use the typedef order on the struct it doesn't work.

dralley:
@kontura Ready for another round of review.  I added a rough cut of the Python API code but it's probably somewhat terrible.

Will work on the tests next.

src/xml_parser_main_metadata_together.c
```diff
@@ -52,30 +77,19 @@ call_user_callback_if_package_finished(cr_Package *pkg, cr_CbData *cb_data, GErr
             g_hash_table_remove(cb_data->in_progress_pkgs_hash, pkg->pkgId);
         } else {
             //remove first element in the list
-            cb_data->in_progress_pkgs_list = cb_data->in_progress_pkgs_list->next;
+            g_slist_delete_link(cb_data->in_progress_pkgs_list, cb_data->in_progress_pkgs_list);
```
kontura, line 80: I think we have to also assign the new head of the list.
```suggestion
            cb_data->in_progress_pkgs_list = g_slist_delete_link(cb_data->in_progress_pkgs_list, cb_data->in_progress_pkgs_list);
```

src/xml_parser_main_metadata_together.c
```diff
@@ -337,133 +351,215 @@ int cr_xml_parse_main_metadata_together(const char *primary_path,
...
-    cr_CbData cbdata;
-    cbdata.in_progress_pkgs_list = NULL;
-    cbdata.in_progress_pkgs_hash = NULL;
-    cbdata.newpkgcb = newpkgcb;
-    cbdata.newpkgcb_data = newpkgcb_data;
-    cbdata.pkgcb = pkgcb;
-    cbdata.pkgcb_data = pkgcb_data;
+    // TODO: do this more properly
+    assert(pkg_iterator);
+    assert(!*err);
```
kontura, line 361: I would prefer if instead of the `assert(!*err);` we return the `err` code, perhaps:
```suggestion
    if (*err) {
        return (*err)->code;
    }
    assert(pkg_iterator);
```
I think there already are some unit tests that need this.

src/xml_parser_main_metadata_together.c
```diff
@@ -337,133 +351,215 @@ int cr_xml_parse_main_metadata_together(const char *primary_path,
...
+            // cr_package_free(package);   // breaks test_xml_parser_main_metadata_together_pkgcb_abort
+            cr_PkgIterator_free(pkg_iterator, err);
+            return (*err)->code;
+        } else {
+            // If callback return CRE_OK but it simultaneously set
+            // the tmp_err then it's a programming error.
+            assert(tmp_err == NULL);
+        };
+    }
+    assert(cr_PkgIterator_is_finished(pkg_iterator) || !*err);
```
kontura, line 383: Same here. I think we can easily run into the assert when there is some problem during parsing for example.
```suggestion
    if (*err) {
        return (*err)->code;
    }
    assert(cr_PkgIterator_is_finished(pkg_iterator));
```

dralley: We're already checking this two lines down, do we need both?  I will go ahead and remove the `!*err` from the assert however

src/python/xml_parser-py.c
```diff
@@ -792,3 +792,248 @@ py_xml_parse_main_metadata_together(G_GNUC_UNUSED PyObject *self, PyObject *args
...
+    cr_XmlParserNewPkgCb ptr_c_newpkgcb = NULL;
+    cr_XmlParserWarningCb ptr_c_warningcb = NULL;
+
+    if (py_newpkgcb != Py_None)
+        ptr_c_newpkgcb = c_newpkgcb;
+    if (py_warningcb != Py_None)
+        ptr_c_warningcb = c_warningcb;
+
+    cbdata.py_newpkgcb = py_newpkgcb;
+    cbdata.py_pkgcb = NULL;  // TODO: What should go here?
```
dralley, line 905: @kontura ^

tests/test_xml_parser_main_metadata_together.c
```diff
@@ -262,37 +263,232 @@ test_cr_xml_parse_main_metadata_together_08_long_primary(void)
...
+        cr_package_free(package);
+    }
+    cr_PkgIterator_free(pkg_iterator, &tmp_err);
+
+    g_assert(tmp_err == NULL);
+    g_assert_cmpint(parsed, ==, 2);
+}
+
+// static void
+// test_cr_xml_parse_main_metadata_together_03_newpkgcb_returns_null(void)
```
dralley, line 327: Which of these tests do you think needs to be duplicated for package iterator, since the `parse_xml_together` tests give us some coverage of it already?  And any others else that might be missing?

dralley:
I dropped the other two commits for now, I'll bring them back once we've gotten the C API portions basically done.

tests/test_xml_parser_main_metadata_together.c
```diff
@@ -262,37 +263,232 @@ test_cr_xml_parse_main_metadata_together_08_long_primary(void)
...
 main(int argc, char *argv[])
 {
     g_test_init(&argc, &argv, NULL);
 
-    g_test_add_func("/xml_parser_filelists/test_cr_xml_parse_main_metadata_together_00",
+    g_test_add_func("/xml_parser_main_metadata/test_cr_xml_parse_main_metadata_together_00",
                     test_cr_xml_parse_main_metadata_together_00);
 
-    g_test_add_func("/xml_parser_filelists/test_cr_xml_parse_main_metadata_together_01_out_of_order_pkgs",
+    // LEAKS
```
dralley, line 451: These are all passing, but a few of them do leak memory atm.

tests/test_xml_parser_main_metadata_together.c
kontura: Right I think we don't need to duplicate all the tests, it looks like you already have the most important ones.

I would maybe add a couple of test for invalid input, such as calling `cr_PkgIterator_prase_next` and `cr_PkgIterator_is_finished` with `NULL` and perhaps `cr_PkgIterator_new` with one of the filepaths set to `NULL`?

src/xml_parser_main_metadata_together.c
kontura: Unless we add the error check there as well, some of the tests are failing for me (running into the `assert(cr_PkgIterator_is_finished(pkg_iterator))`):

`test_cr_xml_parse_main_metadata_together_01_out_of_order_pkgs`
`test_cr_xml_parse_main_metadata_together_04_newpkgcb_interrupt`
`test_cr_xml_parse_main_metadata_together_07_warningcb_interrupt`

So I think we want it there to properly report when there is an error/interupt.

src/xml_parser_main_metadata_together.c
kontura: I think we need to move this `assert(pkg_iterator)` under the `err` check otherwise I am getting a failing test:
`test_cr_xml_parse_main_metadata_together_02_invalid_path`

tests/test_xml_parser_main_metadata_together.c
kontura: Interesting some of them are failing for me without the changes I just suggested above.
Are you sure you are testing with the modified library and not the one present on the system? Or given that all of my fails are caused by `asserts` perhaps those could somehow behave differently?

dralley:
Apologies for the lack of activity, I've been pulled in other directions, and we got the current memory usage down quite a bit by restricting the # of changelogs kept so it became a little lower priority.

@kontura How strongly do you feel about keeping the "out_of_order_pkgs" option around?  I feel like it complicates the lifetimes significantly by having multiple codepaths and isn't necessarily providing enough value to be worth it.  Hitting packages out of order would signal that you've made a mistake 99% of the time and the old API would always work in that scenario.

kontura:
> @kontura How strongly do you feel about keeping the "out_of_order_pkgs" option around? I feel like it complicates the lifetimes significantly and isn't necessarily providing enough value to be worth it. Hitting packages out of order would signal that you've made a mistake 99% of the time and the old API would always work in that scenario, so you still have options.

Not strongly at all, I would be fine with dropping it.

dralley:
I think, most likely we will be pursing a different solution to this problem for Pulp.  The object lifetime model becomes so difficult to manage, due to all of the different callbacks and transitions between data structures and and having two different lifetimes - a refcounted Python package object which references the C package object and has a wholly different lifetime - that even if we get this working it would stretch the architecture to a very fragile place that I'm not sure we could feel comfortable shipping.

The real killer is the fact that even if we get it working there might be unfixable (without a major rewrite) issues with the metadata that gets parsed.

https://github.com/rpm-software-management/createrepo_c/issues/305#issuecomment-1041545459

There are just too many repos in the wild right now with this issue, and not even just third party ones, RHEL6 and centos7-opstools repos suffer from this, and the other common repos like Microsoft Azure's RHEL addons repo had this problem until recently: https://github.com/rpm-software-management/createrepo_c/issues/307

We're still evaluating options but I think the approach in this particular PR is dead.

dralley:
I will take another look at finishing this, in case the issues I was beating my head against the keyboard on are resolved now.  Still not sure about the python bindings.  I was running into a lot of issues where createrepo_c tries to free objects that it jointly owns with PyObject, which of course causes the Python objects to be corrupted.  There's no smooth transfer of control from C to Python.

dralley:
I'm finished rebasing, and the C tests do pass, but the Python tests are of course failing again

src/xml_parser_main_metadata_together.c
```diff
@@ -334,136 +328,227 @@ int cr_xml_parse_main_metadata_together(const char *primary_path,
...
+    );
 
-    cr_CbData cbdata;
-    cbdata.in_progress_pkgs_list = NULL;
-    cbdata.in_progress_pkgs_hash = NULL;
-    cbdata.newpkgcb = newpkgcb;
-    cbdata.newpkgcb_data = newpkgcb_data;
-    cbdata.pkgcb = pkgcb;
-    cbdata.pkgcb_data = pkgcb_data;
+    assert(pkg_iterator);
```
kontura, line 338: This `assert` is causing unittest fails, because the error checking is below it.

src/python/xml_parser-py.c
```diff
@@ -792,3 +790,247 @@ py_xml_parse_main_metadata_together(G_GNUC_UNUSED PyObject *self, PyObject *args
...
+             "    Default constructor\n");
+
+static int
+pkg_iterator_init(_PkgIteratorObject *self, PyObject *args, PyObject *kwargs)
+{
+    char *primary_path;
+    char *filelists_path;
+    char *other_path;
+    PyObject *py_newpkgcb, *py_warningcb;
+    CbData cbdata;
```
kontura, line 860: This is likely the biggest source of the segfaults.
This `cbdata` is valid only in the scope of `pkg_iterator_init` but we save a pointer to it in `cr_PkgIterator_new()`.
Since we need it through out the parsing I think we have allocate it on the heap and store a pointer in `_PkgIteratorObject`.
```suggestion
```

src/python/xml_parser-py.c
```diff
@@ -792,3 +790,247 @@ py_xml_parse_main_metadata_together(G_GNUC_UNUSED PyObject *self, PyObject *args
...
+
+static PyObject *
+pkg_iterator_new(PyTypeObject *type,
+            G_GNUC_UNUSED PyObject *args,
+            G_GNUC_UNUSED PyObject *kwds)
+{
+    _PkgIteratorObject *self = (_PkgIteratorObject *)type->tp_alloc(type, 0);
+    if (self) {
+        self->pkg_iterator = NULL;
+    }
```
kontura, line 844:
```suggestion
        self->cbdata = malloc(sizeof(CbData));
    }
```

src/python/xml_parser-py.c
```diff
@@ -792,3 +790,247 @@ py_xml_parse_main_metadata_together(G_GNUC_UNUSED PyObject *self, PyObject *args
...
+        cr_PkgIterator_free(self->pkg_iterator, &tmp_err);
+    // TODO: correct thing to do with error?
+    if (tmp_err) {
+        nice_exception(&tmp_err, NULL);
+        return -1;
+    }
+
+    Py_XINCREF(py_newpkgcb);
+    Py_XINCREF(py_warningcb);
+    Py_XINCREF(cbdata.py_pkgs);
```
kontura, line 891: This line definitely shouldn't be here. The `cbdata.py_pkgs` dictionary is not even created yet and I don't think we need to do this anyway.
```suggestion
```

src/python/xml_parser-py.c
```diff
@@ -792,3 +790,247 @@ py_xml_parse_main_metadata_together(G_GNUC_UNUSED PyObject *self, PyObject *args
...
+
+    if (py_newpkgcb != Py_None)
+        ptr_c_newpkgcb = c_newpkgcb;
+    if (py_warningcb != Py_None)
+        ptr_c_warningcb = c_warningcb;
+
+    cbdata.py_newpkgcb = py_newpkgcb;
+    cbdata.py_pkgcb = NULL;  // TODO: What should go here?
+    cbdata.py_warningcb = py_warningcb;
+    cbdata.py_pkgs = PyDict_New();
```
kontura, line 904: I think `NULL` for the `py_pkgcb` is fine since it is not used.
```suggestion
    self->cbdata->py_newpkgcb = py_newpkgcb;
    self->cbdata->py_pkgcb = NULL;
    self->cbdata->py_warningcb = py_warningcb;
    self->cbdata->py_pkgs = PyDict_New();
```

src/python/xml_parser-py.c
```diff
@@ -792,3 +790,247 @@ py_xml_parse_main_metadata_together(G_GNUC_UNUSED PyObject *self, PyObject *args
 
     Py_RETURN_NONE;
 }
+
+typedef struct {
+    PyObject_HEAD
+    cr_PkgIterator *pkg_iterator;
+} _PkgIteratorObject;
```
kontura, line 797:
```suggestion
    CbData *cbdata;
} _PkgIteratorObject;
```

src/python/xml_parser-py.c
```diff
@@ -792,3 +790,247 @@ py_xml_parse_main_metadata_together(G_GNUC_UNUSED PyObject *self, PyObject *args
...
+    if (py_warningcb != Py_None)
+        ptr_c_warningcb = c_warningcb;
+
+    cbdata.py_newpkgcb = py_newpkgcb;
+    cbdata.py_pkgcb = NULL;  // TODO: What should go here?
+    cbdata.py_warningcb = py_warningcb;
+    cbdata.py_pkgs = PyDict_New();
+
+    self->pkg_iterator = cr_PkgIterator_new(
+        primary_path, filelists_path, other_path, ptr_c_newpkgcb, &cbdata, ptr_c_warningcb, &cbdata, &tmp_err);
```
kontura, line 907:
```suggestion
        primary_path, filelists_path, other_path, ptr_c_newpkgcb, self->cbdata, ptr_c_warningcb, self->cbdata, &tmp_err);
```

src/python/xml_parser-py.c
```diff
@@ -792,3 +790,247 @@ py_xml_parse_main_metadata_together(G_GNUC_UNUSED PyObject *self, PyObject *args
...
+    }
+
+    if (self->pkg_iterator == NULL) {
+        PyErr_SetString(CrErr_Exception, "PkgIterator initialization failed");
+        return -1;
+    }
+
+    Py_XDECREF(py_newpkgcb);
+    Py_XDECREF(py_warningcb);
+    Py_XDECREF(cbdata.py_pkgs);
```
kontura, line 922: We don't want to do it here yet. The callbacks and `py_pkgs` are used through out the parsing. Py_XDECREFing in `pkg_iterator_dealloc` shoud be fine I think.
```suggestion
```

src/python/xml_parser-py.c
```diff
@@ -792,3 +790,247 @@ py_xml_parse_main_metadata_together(G_GNUC_UNUSED PyObject *self, PyObject *args
...
+    return 0;
+}
+
+static void
+pkg_iterator_dealloc(_PkgIteratorObject *self)
+{
+    GError *tmp_err;
+    if (self->pkg_iterator)
+        cr_PkgIterator_free(self->pkg_iterator, &tmp_err);
+    Py_TYPE(self)->tp_free(self);
```
kontura, line 933: We also have to `free` the mallocated `cbdata`.
```suggestion
    if (self->cbdata) {
        Py_XDECREF(self->cbdata->py_newpkgcb);
        Py_XDECREF(self->cbdata->py_warningcb);
        Py_XDECREF(self->cbdata->py_pkgs);

        free(self->cbdata);
    }
    Py_TYPE(self)->tp_free(self);
```

src/python/xml_parser-py.c
```diff
@@ -792,3 +790,247 @@ py_xml_parse_main_metadata_together(G_GNUC_UNUSED PyObject *self, PyObject *args
...
+        return NULL;
+    }
+
+    if (!pkg) {
+        assert(cr_PkgIterator_is_finished(self->pkg_iterator));
+        PyErr_SetNone(PyExc_StopIteration);
+        return NULL;
+    }
+
+    return Object_FromPackage(pkg, 1);
```
kontura, line 990: Finally given that we don't use `c_pkgcb` we should do a similar dance as it does with the `py_pkgs` dictionary.
So that if the user passes in a package via `newpkgcb` he gets back the same one (filled with data).
```suggestion
    PyObject *keyFromPtr = PyLong_FromVoidPtr(pkg);
    PyObject *py_pkg = PyDict_GetItem(self->cbdata->py_pkgs, keyFromPtr);
    if (py_pkg) {
        // Remove pkg from PyDict but keep one reference so its not freed if the
        // user doesn't have any references to the package
        Py_XINCREF(py_pkg);
        PyDict_DelItem(self->cbdata->py_pkgs, keyFromPtr);
    } else {
        // The package was not provided by user in c_newpkgcb,
        // create new python package object
        py_pkg = Object_FromPackage(pkg, 1);
    }
    Py_DECREF(keyFromPtr);
    return py_pkg;
```

tests/python/tests/test_xml_parser.py
```diff
@@ -1156,3 +1170,187 @@ def warningcb(type, msg):
...
+                 (None, '/usr/share/man/', 'super_kernel.8.gz')])
+        self.assertEqual(pkg.changelogs,
+                [('Tomas Mlcoch <tmlcoch@redhat.com> - 6.0.1-1',
+                   1334664000,
+                  '- First release'),
+                 ('Tomas Mlcoch <tmlcoch@redhat.com> - 6.0.1-2',
+                   1334664001,
+                   '- Second release')])
+
+        self.assertIsNone(next(package_iterator))
```
kontura, line 1242: The iterator raises an `StopIteration` I don't think it return `None`. This repeats on a couple places.
```suggestion
        self.assertRaises(StopIteration, next, package_iterator)
```

tests/python/tests/test_xml_parser.py
```diff
@@ -1156,3 +1170,187 @@ def warningcb(type, msg):
...
+            warningcb=warningcb,
+        )
+
+        package_names = ['fake_bash', 'super_kernel']
+        for (pkg, name) in zip(package_iterator, package_names):
+            self.assertEqual(pkg.name, name)
+            # TODO: might be good to test that iter.is_finished() returns false while still processing packages
+
+        self.assertListEqual(warnings, [])
+        self.assertIsNone(next(package_iterator))
```
kontura, line 1261:
```suggestion
        self.assertRaises(StopIteration, next, package_iterator)
```

tests/python/tests/test_xml_parser.py
```diff
@@ -1156,3 +1170,187 @@ def warningcb(type, msg):
...
+        package_iterator = cr.PackageIterator(
+            primary_path=REPO_02_PRIXML, filelists_path=REPO_02_FILXML, other_path=REPO_02_OTHXML,
+            newpkgcb=newpkgcb,
+        )
+
+        packages = list(package_iterator)
+
+        self.assertEqual(len(packages), 1)
+        self.assertEqual(packages[0].name, "fake_bash")
+        self.assertIsNone(next(package_iterator))
```
kontura, line 1278:
```suggestion
        self.assertRaises(StopIteration, next, package_iterator)
```

tests/python/tests/test_xml_parser.py
```diff
@@ -1156,3 +1170,187 @@ def warningcb(type, msg):
...
+            warningcb=warningcb,
+        )
+
+        package_names = ['fake_bash', 'super_kernel']
+        for (pkg, name) in zip(package_iterator, package_names):
+            self.assertEqual(pkg.name, name)
+            # TODO: might be good to test that iter.is_finished() returns false while still processing packages
+
+        self.assertListEqual(warnings, [])
+        self.assertIsNone(next(package_iterator))
```
kontura, line 1297: The `warnings` are not empty the `assert` below is correct.
```suggestion
        self.assertRaises(StopIteration, next, package_iterator)
```

tests/python/tests/test_xml_parser.py
```diff
@@ -1156,3 +1170,187 @@ def warningcb(type, msg):
...
+
+        package_iterator = cr.PackageIterator(
+            primary_path=PRIMARY_ERROR_00_PATH, filelists_path=FILELISTS_ERROR_00_PATH, other_path=OTHER_ERROR_00_PATH,
+        )
+
+        with self.assertRaises(cr.CreaterepoCError) as ctx:
+            packages = list(package_iterator)
+
+        self.assertIsNone(next(package_iterator))
+        self.assertTrue(package_iterator.is_finished())
```
kontura, line 1324: The `package_iterator` already raised an error, calling `next` on it again is not a good idea. 
But maybe we could fix it so that is doesn't crash? Not sure in what state an object should be after it raises an exception.
```suggestion
```

tests/python/tests/test_xml_parser.py
```diff
@@ -1156,3 +1170,187 @@ def warningcb(type, msg):
...
+        package_iterator = cr.PackageIterator(
+            primary_path=REPO_02_PRIXML, filelists_path=REPO_02_FILXML, other_path=REPO_02_OTHXML,
+            newpkgcb=newpkgcb,
+        )
+
+        with self.assertRaises(cr.CreaterepoCError) as ctx:
+            packages = list(package_iterator)
+
+        self.assertIsNone(next(package_iterator))
+        self.assertTrue(package_iterator.is_finished())
```
kontura, line 1340: The `package_iterator` already raised an error.
```suggestion
```

tests/python/tests/test_xml_parser.py
```diff
@@ -1156,3 +1170,187 @@ def warningcb(type, msg):
...
+        package_iterator = cr.PackageIterator(
+            primary_path=PRIMARY_MULTI_WARN_00_PATH, filelists_path=FILELISTS_MULTI_WARN_00_PATH, other_path=OTHER_MULTI_WARN_00_PATH,
+            warningcb=warningcb,
+        )
+
+        with self.assertRaises(cr.CreaterepoCError) as ctx:
+            packages = list(package_iterator)
+
+        self.assertIsNone(next(package_iterator))
+        self.assertTrue(package_iterator.is_finished())
```
kontura, line 1356: The `package_iterator` already raised an error.
```suggestion
```

kontura (changes requested):
I am proposing quite a few changes but unless I missed something while converting this to github review it should work well with them.
I am not seeing any segfauls or memory leaks.
Please let me know what you think. :slightly_smiling_face: 
(just for convenience I have uploaded the changes also here https://github.com/kontura/createrepo_c/tree/pr-295)

dralley:
@kontura Could you paste the patch?  I can just apply it directly to the PR.

dralley:
(If you already have one - I assume you do since you mentioned testing it)

tests/python/tests/test_xml_parser.py
dralley: Correct, I was probably thinking in terms of the C implementation

kontura:
> @kontura Could you paste the patch? I can just apply it directly to the PR.

I have pushed them as commits to my branch: https://github.com/kontura/createrepo_c/commits/pr-295

dralley:
@kontura Thanks!

This does pass all the tests, with no valgrind warnings, despite the fact that some of the code hurts my soul a little bit :)

I need to do some additional testing with my collection of deviant repos to make sure everything is still looking correct, will update later.

dralley:
>>> I hope it will simplify a bit when we switch to the "pull" / "StAX" parsing model.

>>Is this a plan now? :)

> I would definitely like to do it, hopefully in not so distant future. slightly_smiling_face

If that does eventually happen, feel free to copy from my implementation https://github.com/dralley/rpmrepo_metadata/

dralley:
It seems to work fine for all the repos I tested!  That is excellent.  Would you mind if I cleaned up the examples a bit to promote the newer parsing methods?

src/xml_parser_main_metadata_together.c
```diff
@@ -334,136 +328,227 @@ int cr_xml_parse_main_metadata_together(const char *primary_path,
...
+                // the tmp_err then it's a programming error.
+                assert(tmp_err == NULL);
+            }
+        } else {
+            // Free the package if there is no external callback to do so.
+            cr_package_free(package);
+        }
     }
 
+    assert(cr_PkgIterator_is_finished(pkg_iterator));
```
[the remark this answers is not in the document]
dralley: I'm just rebuilding via the Python package, and I believe that always (or at least by default) builds w/ release configuration, so that's probably true.

kontura:
>  Would you mind if I cleaned up the examples a bit to promote the newer parsing methods?

Sure, I would also like to remove the `xml_parse_main_metadata_together(..)` example and even make the API function deprecated. So that we have just one way to parse the files together.

Too bad its already in rhel, otherwise we could just remove it.

dralley:
The test I've just added triggers a segfault

src/python/xml_parser-py.c
```diff
@@ -792,3 +790,271 @@ py_xml_parse_main_metadata_together(G_GNUC_UNUSED PyObject *self, PyObject *args
...
+static void
+pkg_iterator_dealloc(_PkgIteratorObject *self)
+{
+    GError *tmp_err;
+    if (self->pkg_iterator) {
+        cr_PkgIterator_free(self->pkg_iterator, &tmp_err);
+    }
+    if (self->cbdata) {
+        // Py_XDECREF(self->cbdata->py_newpkgcb);
+        // Py_XDECREF(self->cbdata->py_warningcb);
```
dralley, line 937: I've commented out these two lines, uncommenting either of them leads to segfault on the added test

src/xml_parser_main_metadata_together.c
```diff
@@ -334,136 +328,231 @@ int cr_xml_parse_main_metadata_together(const char *primary_path,
...
+
+    if (*err) {
+        return (*err)->code;
+    } else {
+        return CRE_OK;
+    }
+}
+
+// TODO: maybe whether or not individual files are parsed could be controlled by NULL paths? I think cr_load_xml_files
+// already works that way.
```
dralley, line 382: I took a brief crack at this, but lines 466, 476, 486 make it tricky.

dralley:
>>    Would you mind if I cleaned up the examples a bit to promote the newer parsing methods?

> Sure, I would also like to remove the xml_parse_main_metadata_together(..) example and even make the API function deprecated. So that we have just one way to parse the files together.

> Too bad its already in rhel, otherwise we could just remove it.

Considering the change to remove `allow_out_of_order_pkgs` is both ABI and API incompatible, are we sure about that?

kontura:
> Also the change to remove allow_out_of_order_pkgs is both ABI and API incompatible, if that's not a blocker then I wouldn't think it would be a blocker to remove the interface entirely especially given how new it is.

I think it would be a problem.

> Is it though? It's only in RHEL9 Appstream which I would assume isn't considered 100% locked down yet.

Right I was looking at our rhel distgits where the changes are already present but yea they are not released yet. I will look into what could we do.

dralley:
Spent another hour or so on trying to make it capable of taking NULL arguments, and for the sake of actually getting this merged, I think it's best that we don't include that functionality for now.

dralley:
Spent another hour or so on trying to make it capable of taking NULL arguments, and for the sake of actually getting this merged, I'm thinking it's best that we don't include that functionality for now.

src/python/xml_parser-py.c
```diff
@@ -790,3 +790,241 @@ py_xml_parse_main_metadata_together(G_GNUC_UNUSED PyObject *self, PyObject *args
...
+        return -1;
+    }
+
+    if (self->pkg_iterator) // reinitialization by __init__()
+        cr_PkgIterator_free(self->pkg_iterator, &tmp_err);
+    // TODO: correct thing to do with error?
+    if (tmp_err) {
+        nice_exception(&tmp_err, NULL);
+        return -1;
+    }
```
dralley, line 893: Correct?

kontura: I think so.
The `nice_exception()` helper handles the `tmp_err` and calls `PyErr_SetString()` :+1: .

src/python/xml_parser-py.c
kontura: I see, `Py_XDECREF` can take `NULL`s so the problem is we don't always initialize `self->cbdata` it is just `malloced` so there are random data (random addresses) and that doesn't go well.

I think the best fix would be to change the line 
```c
self->cbdata = malloc(sizeof(CbData));
```
to
```c
self->cbdata = g_malloc0(sizeof(CbData));
```

kontura (changes requested):
Could you also clean up the commits? I would prefer if you squashed the fix up commits so that we have a cleaner history of just commits adding the new API.

src/python/xml_parser-py.c
```diff
@@ -790,3 +790,240 @@ py_xml_parse_main_metadata_together(G_GNUC_UNUSED PyObject *self, PyObject *args
...
+{
+    char *primary_path;
+    char *filelists_path;
+    char *other_path;
+    PyObject *py_newpkgcb, *py_warningcb;
+    GError *tmp_err = NULL;
+    static char *kwlist[] = {"primary", "filelists", "other", "newpkgcb",
+                             "warningcb", NULL};
+
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "sssOO:pkg_iterator_init", kwlist,
```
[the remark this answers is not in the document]
dralley: It's better as it is now if the underlying code can't handle it yet

src/python/xml_parser-py.c
```diff
@@ -790,3 +790,238 @@ py_xml_parse_main_metadata_together(G_GNUC_UNUSED PyObject *self, PyObject *args
...
+    cr_XmlParserNewPkgCb ptr_c_newpkgcb = NULL;
+    cr_XmlParserWarningCb ptr_c_warningcb = NULL;
+
+    if (py_newpkgcb != Py_None)
+        ptr_c_newpkgcb = c_newpkgcb;
+    if (py_warningcb != Py_None)
+        ptr_c_warningcb = c_warningcb;
+
+    self->cbdata->py_newpkgcb = py_newpkgcb;
+    self->cbdata->py_pkgcb = NULL;  // TODO: What should go here?
```
dralley, line 906: Can I remove the comment?

src/python/xml_parser-py.c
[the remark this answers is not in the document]
dralley: Although, something is off still

```
TypeError: pkg_iterator_init() argument 1 must be str, not None

In [3]: exit
Segmentation fault (core dumped)
```

src/python/xml_parser-py.c
kontura: Sure, no problem.

src/python/xml_parser-py.c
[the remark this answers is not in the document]
kontura: I think its still this issue: https://github.com/rpm-software-management/createrepo_c/pull/295#discussion_r823429255

If there is an error we return before `cbdata` is initialized and when `pkg_iterator_dealloc()` is called we run `PY_XDECRED()` on random values.

kontura:
Thank you @dralley for all the hard work!

We have also managed to not release the obsoleted API (`cr_xml_parse_main_metadata_together`) so I will make a follow up PR to remove it.

PR #296, opened 2021-11-28
Fix a bug in cr_repomd_record_compress_and_fill()

dralley:
These are the two changes suggested here: https://github.com/rpm-software-management/createrepo_c/pull/271/files#r703194629


dralley:
I haven't done any special testing or tried to reproduce the issue, FYI.

kontura:
Thanks, I have completely forgotten about this.

PR #304, opened 2022-02-02
Make parse warnings visible through cr.Repomd(), cr.UpdateInfo()

dralley:
There is no way to access parser warnings from the higher-level API.

### DIFF
--- a/src/python/createrepo_c/__init__.py
+++ b/src/python/createrepo_c/__init__.py
@@ -149,8 +149,14 @@ class Repomd(_createrepo_c.Repomd):
     def __init__(self, path=None):
         """:arg path: Path to existing repomd.xml or None"""
         _createrepo_c.Repomd.__init__(self)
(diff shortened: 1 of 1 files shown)


src/python/createrepo_c/__init__.py
```diff
@@ -145,12 +145,19 @@ def __init__(self, filename, mode=MODE_READ,
 
 # Repomd class
 
+
 class Repomd(_createrepo_c.Repomd):
     def __init__(self, path=None):
```
dralley, line 150: The other option would be to expose the `warningcb` directly.  That felt a little awkward since it would mix API styles (it's not the way you would typically do that sort of thing in Python) but I wouldn't have a problem with doing it that way if it's what you would prefer.

kontura: I think this is fine, thanks!

Issue #305, opened 2022-02-10
Segmentation fault when parsing repo metadata


dralley:
Only applies to the new `parse_main_metadata_together` API, and only when parsing this repository: http://software.internet2.edu/rpms/el7/x86_64/main/

dralley:
This repo has every package listed twice (same pkgid, same nevra x2), I've already reported that to the distributor so hopefully they will address it.  Although this segfault still needs to be fixed :)

dralley:
```
==333342== Invalid write of size 8                                                                                                                                                                                                            
==333342==    at 0x6D0C323: g_string_chunk_insert_len (gstringchunk.c:297)                                                                                                                                                                    
==333342==    by 0x6BC6221: cr_end_handler (in /usr/local/lib/pulp/lib/python3.10/site-packages/createrepo_c/_createrepo_c.so)                                                                                                                
==333342==    by 0x701B734: xmlParseEndTag1.constprop.0 (parser.c:8708)                                                                                                                                                                       
==333342==    by 0x70244AE: xmlParseTryOrFinish.isra.0 (parser.c:11642)                                                                                                                                                                       
==333342==    by 0x6F403A8: xmlParseChunk (parser.c:12346)                                                                                                                                                                                    
==333342==    by 0x6BCA2A7: parse_next_section (in /usr/local/lib/pulp/lib/python3.10/site-packages/createrepo_c/_createrepo_c.so)                                                                                                            
==333342==    by 0x6BCB060: cr_xml_parse_main_metadata_together (in /usr/local/lib/pulp/lib/python3.10/site-packages/createrepo_c/_createrepo_c.so)                                                                                           
==333342==    by 0x6BB52B5: py_xml_parse_main_metadata_together (in /usr/local/lib/pulp/lib/python3.10/site-packages/createrepo_c/_createrepo_c.so)                                                                                           
==333342==    by 0x4978F01: cfunction_call (methodobject.c:543)                                                                                                                                                                               
==333342==    by 0x4974BD3: _PyObject_MakeTpCall (call.c:215)                                                                                                                                                                                 
==333342==    by 0x49719BD: UnknownInlinedFun (abstract.h:112)                                                                                                                                                                                
==333342==    by 0x49719BD: UnknownInlinedFun (abstract.h:99)                                                                                                                                                                                 
==333342==    by 0x49719BD: UnknownInlinedFun (abstract.h:123)                                                                                                                                                                                
[...]

PR #311, opened 2022-02-22
Remove allow_out_of_order option and fix couple of leaks

kontura:
The option was causing issues when repository has duplicate packages and
it was providing little value. All repos should really have the same
order of packages in primary, filelists and other metadata.

For: https://github.com/rpm-software-management/createrepo_c/issues/305

It also includes 2 fixes for memory leaks originally from https://github.com/rpm-software-management/createrepo_c/pull/295

### DIFF
--- a/src/xml_parser_main_metadata_together.c
+++ b/src/xml_parser_main_metadata_together.c
@@ -462,7 +462,11 @@ out:
     if (allow_out_of_order) {
         g_hash_table_destroy(cbdata.in_progress_pkgs_hash);
     } else {
-        cr_slist_free_full(cbdata.in_progress_pkgs_list, (GDestroyNotify) cr_package_free);
+        if (cbdata.newpkgcb) {
+            g_slist_free(cbdata.in_progress_pkgs_list);
+        } else {
+            cr_slist_free_full(cbdata.in_progress_pkgs_list, (GDestroyNotify) cr_package_free);
+        }
     }
 
     return ret;
--- a/src/xml_parser_main_metadata_together.c
+++ b/src/xml_parser_main_metadata_together.c
@@ -432,15 +432,27 @@ out:
         }
     }
 
-    cr_xml_parser_data_free(primary_pd);
-    cr_xml_parser_data_free(filelists_pd);
-    cr_xml_parser_data_free(other_pd);
+    // When interrupted at the right time primary_pd->pkg can either be:
+    //  - referenced in both primary_pd->pkg and cbdata.in_progress_pkgs_list or
+    //  - referenced only in primary_pd->pkg (we have started parsing the pkg but didn't finish it yet)
+    // in order to avoid a crash remove it from the list if present
+    // primary_pd->pkg is a special case because of newpkgcb_primary
+    if (cbdata.in_progress_pkgs_list) {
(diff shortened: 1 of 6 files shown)


src/xml_parser_main_metadata_together.c
```diff
@@ -432,15 +432,27 @@ int cr_xml_parse_main_metadata_together(const char *primary_path,
...
+    //  - referenced in both primary_pd->pkg and cbdata.in_progress_pkgs_list or
+    //  - referenced only in primary_pd->pkg (we have started parsing the pkg but didn't finish it yet)
+    // in order to avoid a crash remove it from the list if present
+    // primary_pd->pkg is a special case because of newpkgcb_primary
+    if (cbdata.in_progress_pkgs_list) {
+        cbdata.in_progress_pkgs_list = g_slist_remove(cbdata.in_progress_pkgs_list, primary_pd->pkg);
+    }
+    if (primary_pd) {
+        cr_package_free(primary_pd->pkg);
+    }
```
dralley, line 445: IIRC (it has been a month since I last looked at all of this) that it gets more complicated when the Python bindings are involved, because there can be other indirect references via https://github.com/rpm-software-management/createrepo_c/blob/master/src/python/xml_parser-py.c#L37 and via the PyObjects themselves, and having the C code free the package can lead to problems when the package PyObjects get DECREF'd and try to free the package payload (again).

src/xml_parser_main_metadata_together.c
```diff
@@ -462,7 +462,11 @@ int cr_xml_parse_main_metadata_together(const char *primary_path,
     if (allow_out_of_order) {
         g_hash_table_destroy(cbdata.in_progress_pkgs_hash);
     } else {
-        cr_slist_free_full(cbdata.in_progress_pkgs_list, (GDestroyNotify) cr_package_free);
+        if (cbdata.newpkgcb) {
+            g_slist_free(cbdata.in_progress_pkgs_list);
+        } else {
+            cr_slist_free_full(cbdata.in_progress_pkgs_list, (GDestroyNotify) cr_package_free);
+        }
```
dralley: IIRC - and it's been a while since I last looked at this - my struggle was that I'm not sure this would work for anyone using the C API directly, since they wouldn't have the Python layer to de-allocate the packages for them.

dralley:
@kontura I've rebased https://github.com/rpm-software-management/createrepo_c/pull/295 on top of these commits and while the situation does seem improved I'm still seeing a bunch of segfaults, would you mind taking a look?

src/xml_parser_main_metadata_together.c
kontura: Right, I think that is fine though.

The way I see it the user (this can be our Python bindings or any other C API direct user) is responsible for managing the packages if they created them.

I think it wouldn't be nice to free a package that was passed to `createrepo_c` in a callback, the user (even C API user) can still have a reference to it. In fact they should have it I think that is the main use case for `newpkgcb`.

I will add a note about this to the doc string.

dralley:
FWIW, this PR on its own seems to work perfectly fine

Conan-Kudo (approved):

PR #316, opened 2022-03-24
Add optional filelists_ext metadata

### LINKED ISSUE #313: IMA: Extend filelists.xml to include a hash attribute inside the `file` elements.
Recently we are looking for ways to securely deliver the file hashes
of the installed RPMs, without downloading the RPMs and inspecting
them locally.

The repository metadata contains in the repomd.xml the list of the
different files that compose the repository, and this file can be
currently signed and validated.  From here, we can also validate the
different contents of the "primary", "filelists" and "other" XML,
calculating the SHA256 and comparing it with the prefix of the file
name.

This makes a perfect channel to deliver validated information about
the content of the RPMs, without requiring the direct download,
validation and inspection of one and each of the RPMs inside the
repository.  This is currently done for the file list, the changelog,
requirement or recommendation of the packages.

Can be very interesting if we extend "filelists.xml" to include one
attribute into the `file` element, that currently enumerate the
different files that belong to a RPM.  This attribute will be `hash`,
and will contain the user defined hash (maybe sha256) of each file.

For example, currently a `file` element is like:

   `<file>/usr/bin/bash</file>`

I propose to be extended as:

   `<file hash="7d301d1f90ac6e56f171d5de459b887eb6077c6e0ea340c04fe23855b1c6b9a3">/usr/bin/bash</file>`

The hash function can be selected by the user with a parameter during
the `createrepo` call.

  `createrepo --hash sha256`

Because currently the RPM header is already containing a hash for each
file, could make sense to use directly this value, making the hash
funcion parameter useless.

  `createrepo --hash`

One use of this new feature is for remote attestation.  In remote
attestation (for example, using services like Keylime), we can enable
the IMA logging in the kernel of a monitored system, that will
routinely send all the IMA hashes of the opened or executed files.

The remote node will compare those IMA hashes registered by the kernel
and compare it with a white list.  One problem is how to generate this
list without accessing the monitored system.  With this proposed
feature the white list can be generated knowing only the list of
repositories used during the installation.

aplanas:
Using the parameter '--filelists_ext', a new metadata file
(filelists_ext) will be created.  This new metadata is a copy of
filelists, but will include a new attribute for the 'file' element,
named 'hash', that contain the registered hash (SHA256 for example)
for this file.

Fix #313

Signed-off-by: Alberto Planas <aplanas@suse.com>


aplanas:
There are some bits missing, mostly tests, some adjustments in the Python bindings, and all the sqlite code.

But I want to present this patch as early as possible, to see if the approach will be OK

mlschroe:
No testcases?

aplanas:
> The name "--hash" is somewhat confusing.

Done. I renamed it to `--filelists_ext`

kontura:
Given that the `filelists-ext` fully contains regular `filelists` how about instead of adding all the new parsing APIs we just use the existing ones? I think it would just work since the element was added to `xml_parser_filelists.c`.

src/xml_parser_filelists.c
```diff
@@ -195,6 +210,9 @@ cr_start_handler(void *pdata, const xmlChar *element, const xmlChar **attr)
                 cr_xml_parser_warning(pd, CR_XML_WARNING_UNKNOWNVAL,
                                       "Unknown file type \"%s\"", val);
         }
+
+        if (!pd->last_digest)
+            pd->last_digest = cr_find_attr("hash", attr);
```
kontura, line 215: I think you have to make a copy here, the `attr` don't live beyond the `cr_start_handler`.

aplanas:
> Given that the `filelists-ext` fully contains regular `filelists` how about instead of adding all the new parsing APIs we just use the existing ones? I think it would just work since the element was added to `xml_parser_filelists.c`.

I like the idea. The parser state machine is shared, I will see how I can unify the API. I must admit that when I do not understand the code, I make an explicit copy for the filelists_ext version.

src/xml_parser_filelists.c
aplanas: > What should I be doing?

Ok, I found a similar case with `cpid` field. I will use the same approach/

aplanas: Done. That was not the coredump : (

kontura: If I am not mistaken you can add it right in to the `pd->pkg->chunk` and then in the `cr_end_handler` just copy the pointer.
You also should reset the `pd->last_digest` or don't check here if it is already set otherwise you will set it only once and all the files will have the same digest.

aplanas:
I still plan to add some tests, but I am thinking to drop the changes in sqlite for now. Meanwhile I moved the PR to ready for reviews.

aplanas:
I updated some tests, and I decided not to break the cr_PackageFiles Python tuple, and not include the digest there.

IMHO I am done.

dralley:
> I still plan to add some tests, but I am thinking to drop the changes in sqlite for now. Meanwhile I moved the PR to ready for reviews.

Yes, what I've now realized is that .sqlite metadata is de-facto deprecated even though it's still provided in most repos.  None of the tools that consume it are mainstream anymore, except for legacy `yum` in el7, which has 18 months to live.

I filed this issue to turn it off in a future Fedora release.  https://pagure.io/releng/issue/10745

So no changes to the sqlite metadata is likely preferable.

src/xml_dump.c
```diff
@@ -195,6 +195,10 @@ cr_xml_dump_files(xmlNodePtr node, cr_Package *package, int primary)
         if (entry->type && entry->type[0] != '\0' && strcmp(entry->type, "file")) {
             cr_xmlNewProp(file_node, BAD_CAST "type", BAD_CAST entry->type);
         }
+
+        if (filelists_ext && entry->digest && entry->digest[0] != '\0') {
+            cr_xmlNewProp(file_node, BAD_CAST "hash", BAD_CAST entry->digest);
```
dralley, line 200: This probably should have an attribute to distinguish the digest algorithm used alongside the digest itself.

tests/test_xml_parser_main_metadata_together.c
```diff
@@ -120,6 +120,28 @@ test_cr_xml_package_iterator_00(void)
...
 
+static void
+test_cr_xml_package_iterator_filelists_ext_00(void)
+{
+    int parsed = 0;
+    GError *tmp_err = NULL;
+    cr_Package *package = NULL;
+
+    cr_PkgIterator *pkg_iterator = cr_PkgIterator_new(
+        TEST_REPO_04_PRIMARY, TEST_REPO_04_FILELISTS_EXT, TEST_REPO_04_OTHER, NULL, NULL, NULL, NULL, &tmp_err);
```
dralley, line 131: My understanding of this API then, is that if you want the full metadata, you substitute the path to `filelists_ext` instead of `filelists`?

aplanas: For the parser, this is the idea.

src/xml_dump.c
aplanas: There is a new XML field for each package, that indicate the algorithm (`<checksum/>`), this tag is the file digest itself

Conan-Kudo:
@aplanas Are you still interested in this? It needs rebasing...

aplanas:
@Conan-Kudo Yes. Rebased!

Is there anything actionable that I can do to bootstrap the discussion? Is the goal of this PR clear? Maybe there is a better way to connect the signature of the entity that create the repository with the value of the hashes?

aplanas:
@Conan-Kudo ping. Can you remove the "decaying" tag?

src/parsehdr.c
```diff
@@ -298,6 +336,7 @@ cr_package_from_header(Header hdr,
         rpmtdFreeData(filenames);
         rpmtdFreeData(fileflags);
         rpmtdFreeData(filemodes);
+        rpmtdFreeData(filedigests);
     }
 
     rpmtdFree(dirnames);
```
kontura, line 342: I think there is a memory leak of the `filedigests`.
I believe we should also do `rpmtdFree(filedigests);` here.

src/parsepkg.h
```diff
@@ -99,6 +100,7 @@ struct cr_XmlStruct cr_xml_from_rpm(const char *filename,
                                     const char *location_href,
                                     const char *location_base,
                                     int changelog_limit,
+                                    gboolean filelists_ext,
```
kontura, line 103: Could you create a new function instead of changing this one?

src/xml_dump.h
```diff
@@ -104,19 +107,27 @@ char *cr_xml_dump_primary(cr_Package *package, GError **err);
...
 
-/** Generate all three xml chunks (primary, filelists, other) from cr_Package.
+/** Generate all three xml chunks (primary, filelists[_ext], other) from cr_Package.
  * @param package       cr_Package
+ * @param filelists_ext gboolean
  * @param err           **GError
  * @return              cr_XmlStruct
  */
-struct cr_XmlStruct cr_xml_dump(cr_Package *package, GError **err);
+struct cr_XmlStruct cr_xml_dump(cr_Package *package, gboolean filelists_ext, GError **err);
```
kontura, line 130: Please create a new function.

kontura (changes requested):
Hello, sorry for the delay.

Unless I am wrong this breaks API only with the two functions I annotated, please can you rather create new ones? (the Python API should be fine as the new params are optional).

Otherwise in my opinion this is good to go and I would merge it.

src/parsehdr.c
aplanas: yes, done

src/parsepkg.h
aplanas: yes, done, but in the Python side I kept one, as it is an optional parameter that does not break the old signature function

src/xml_dump.h
aplanas: done (also not done in the Python side for the same reason)

aplanas:
Another option instead to duplicate similar functions, can be to drop --filelist_ext and generate it always, and is the user who decides to ignore it or not

dralley:
You probably don't need to "duplicate" per se, just have one implementation with a different signature which is called by the existing function, which would be turned into just a stub.

aplanas:
> just have one implementation with a different signature which is called by the existing function

Right, I used this approach.

kontura (approved):

PR #322, opened 2022-06-22
createrepo_c shouldn't silently produce duplicate-NEVRA repos

### LINKED ISSUE #307: createrepo_c does not warn or error when building a repository where the same package is present more than once
My understanding is that "pkgid" and package NEVRA are meant to be unique in a repo.  createrepo_c (the command line tool, at least) ought to provide a warning or error when commanded to build a repo with the same package present twice

This can occur when:

* The user thinks they can just rename the RPM to make it a different package (the package metadata is in the header so obviously this doesn't work)
* The user puts the same RPM in multiple sub-directories

This is frequently seen in the wild even in commonly used community repos (centos7-opstools) and professional repos (prometheus, gitlab, perf-sonar, RHEL6, microsoft azure RHEL additions)

dralley:
Return an error code and print a message when more than one package have
the same NEVRA.

closes #307

### DIFF
--- a/src/createrepo_c.c
+++ b/src/createrepo_c.c
@@ -553,6 +553,31 @@ load_old_metadata(cr_Metadata **md,
               g_hash_table_size(cr_metadata_hashtable(*md)));
 }
 
+// Sorting function for location_href strings, by length.
+// Compatible with g_array_sort()
+static int strlensort(gconstpointer a, gconstpointer b)
+{
+    // Function is supposed to take a double-pointer so unfortunately you cannot pass a
+    // string-comparison function directly.
+    gchar **a_ptr = (gchar **)a;
+    gchar **b_ptr = (gchar **)b;
+
+    int a_len = strnlen(*a_ptr, 4096);
+    int b_len = strnlen(*b_ptr, 4096);
+    if (a_len > b_len)
+    {
+        return 1;
+    }
+    else if (b_len > a_len)
+    {
+        return -1;
+    }
+    else
+    {
+        return 0;
+    }
+}
+
 int
 main(int argc, char **argv)
 {
@@ -1276,6 +1301,7 @@ main(int argc, char **argv)
     user_data.repodir_name_len  = strlen(in_dir);
     user_data.task_count        = task_count;
     user_data.package_count     = 0;
+    user_data.nevra_table       = g_hash_table_new(g_str_hash, g_str_equal);
     user_data.skip_stat         = cmd_options->skip_stat;
     user_data.old_metadata      = old_metadata;
     user_data.id_pri            = 0;
@@ -1290,6 +1316,7 @@ main(int argc, char **argv)
     user_data.had_errors        = 0;
     user_data.output_pkg_list   = output_pkg_list;
 
+    g_mutex_init(&(user_data.mutex_nevra_table));
     g_mutex_init(&(user_data.mutex_output_pkg_list));
     g_mutex_init(&(user_data.mutex_pri));
     g_mutex_init(&(user_data.mutex_fil));
@@ -1310,9 +1337,36 @@ main(int argc, char **argv)
     // Wait until pool is finished
     g_thread_pool_free(pool, FALSE, TRUE);
 
+    GHashTableIter iter;
+    gpointer key, value;
+
+    g_hash_table_iter_init(&iter, user_data.nevra_table);
+    while (g_hash_table_iter_next(&iter, &key, &value))
+    {
+        gchar *nevra = (gchar *) key;
+        GArray *locations = (GArray *) value;
+        if (locations->len > 1) {
+            g_warning("Package '%s' has duplicate metadata entries, only one should exist", nevra);
+
+            g_array_sort(locations, strlensort);
+
+            for (int i=0; i<locations->len; i++) {
+                g_warning("    Sourced from location: \'%s\'", g_array_index(locations, gchar *, i));
+            }
+        }
+
+        g_hash_table_iter_steal(&iter);
+        g_free(nevra);
+        for (int i = 0; i < locations->len; i++) {
+            g_free(g_array_index(locations, gchar *, i));
+        }
+        g_array_free(locations, TRUE);
+    }
+    g_hash_table_destroy(user_data.nevra_table);
+
     // if there were any errors, exit nonzero
     if ( user_data.had_errors ) {
-	exit_val = 2;
+	    exit_val = 2;
     }
 
     g_message("Pool finished%s", (user_data.had_errors ? " with errors" : ""));
@@ -1453,6 +1507,7 @@ main(int argc, char **argv)
     }
 
     g_queue_free(user_data.buffer);
+    g_mutex_clear(&(user_data.mutex_nevra_table));
     g_mutex_clear(&(user_data.mutex_output_pkg_list));
     g_mutex_clear(&(user_data.mutex_pri));
     g_mutex_clear(&(user_data.mutex_fil));
--- a/src/dumper_thread.c
+++ b/src/dumper_thread.c
@@ -551,6 +551,19 @@ cr_dumper_thread(gpointer data, gpointer user_data)
     }
 #endif
 
+    // Allow checking that the same package (NEVRA) isn't present multiple times in the metadata
+    // Keep a hashtable of NEVRA mapped to an array-list of location_href values
+    g_mutex_lock(&(udata->mutex_nevra_table));
+    gchar *nevra = cr_package_nevra(pkg);
+    GArray *pkg_locations = g_hash_table_lookup(udata->nevra_table, nevra);
+    if (!pkg_locations) {
+        pkg_locations = g_array_new(FALSE, TRUE, sizeof(gchar *));
+        g_hash_table_insert(udata->nevra_table, nevra, pkg_locations);
+    }
+    gchar *location = g_strdup(pkg->location_href);
+    g_array_append_val(pkg_locations, location);
+    g_mutex_unlock(&(udata->mutex_nevra_table));
+
     // Buffering stuff
     g_mutex_lock(&(udata->mutex_buffer));
 
--- a/src/dumper_thread.h
+++ b/src/dumper_thread.h
@@ -69,6 +69,10 @@ struct UserData {
     long task_count;                // Total number of task to process
     long package_count;             // Total number of packages processed
 
+    // Duplicate package error checking
+    GMutex mutex_nevra_table;       // Mutex for the table of NEVRAs
+    GHashTable *nevra_table;         // Table of NEVRAs, with a list of location_href as key
+
     // Update stuff
     gboolean skip_stat;             // Skip stat() while updating
     cr_Metadata *old_metadata;      // Loaded metadata


src/createrepo_c.c
```diff
@@ -427,10 +427,10 @@ cr_create_repomd_records_for_groupfile_metadata(const cr_Metadatum *group_metada
 }
 
 /** Creates list of cr_RepomdRecords from list
- *  of additional metadata (cr_Metadatum) 
+ *  of additional metadata (cr_Metadatum)
  *
  * @param additional_metadata       List of cr_Metadatum
- * @param repomd_checksum_type
```
dralley, line 433: Sorry about this, my editor did it, I can get rid of these changes if you want.

src/createrepo_c.c
```diff
@@ -1306,9 +1308,34 @@ main(int argc, char **argv)
...
+    GHashTableIter iter;
+    gpointer key, value;
+
+    g_hash_table_iter_init(&iter, user_data.nevra_table);
+    while (g_hash_table_iter_next(&iter, &key, &value))
+    {
+        gchar *nevra = (gchar *) key;
+        GArray *locations = (GArray *) value;
+        if (locations->len > 1) {
+            g_warning("Package '%s' has duplicate metadata entries, only one should exist", nevra);
```
dralley, line 1320: `g_error()` makes it terminate immediately, which I would be somewhat fine with, except that it prevents logging information on how to fix the problem.

dralley:
```
(.env) [dalley@thinkpad test]$ createrepo_c .
Directory walk started
Directory walk done - 3 packages
Temporary output repo path: ./.repodata/
Preparing sqlite DBs
Pool started (with 5 workers)
Warning: Package 'wget-0:1.21.3-1.fc35.x86_64' has duplicate metadata entries, only one should exist
Warning:     Sourced from location: 'bar/wget-1.21.3-1.fc35.x86_64.rpm'
Warning:     Sourced from location: 'foo/wget-1.21.3-1.fc35.x86_64.rpm'
Warning:     Sourced from location: 'wget-1.21.3-1.fc35.x86_64.rpm'

```

src/createrepo_c.c
kontura: I think its not a problem here.

src/createrepo_c.c
```diff
@@ -1286,6 +1287,7 @@ main(int argc, char **argv)
     user_data.had_errors        = 0;
     user_data.output_pkg_list   = output_pkg_list;
 
+    g_mutex_init(&(user_data.mutex_nevra_table));
```
kontura, line 1290: We have to also `g_mutex_clear()` the `user_data.mutex_nevra_table`, its around [here](https://github.com/rpm-software-management/createrepo_c/blob/master/src/createrepo_c.c#L1452-L1455).

Conan-Kudo:
This will break COPR. cc: @praiskup

kontura (changes requested):
I have also created CI test update: https://github.com/rpm-software-management/ci-dnf-stack/pull/1129

src/createrepo_c.c
dralley: @kontura Done

dralley:
@kontura Left a comment here https://github.com/rpm-software-management/ci-dnf-stack/pull/1129/files#r903764740=

dralley:
Would this be unreasonable to backport to, say, EL8 and EL9?  (or not backport, but rather, rebase to the latest available)

src/createrepo_c.c
```diff
@@ -1306,9 +1308,51 @@ main(int argc, char **argv)
...
+                if (a_len > b_len) {
+                    return 1;
+                } else if (b_len > a_len) {
+                    return -1;
+                } else {
+                    return 0;
+                }
+            }
+
+            g_array_sort(locations, strsort);
```
dralley, line 1337:
```
(.env) [dalley@thinkpad test]$ createrepo_c .
Directory walk started
Directory walk done - 5 packages
Temporary output repo path: ./.repodata/
Preparing sqlite DBs
Pool started (with 5 workers)
Warning: Package 'htop-0:3.2.1-1.fc35.x86_64' has duplicate metadata entries, only one should exist
Warning:     Sourced from location: 'htop-3.2.1-1.fc35.x86_64.rpm'
Warning:     Sourced from location: 'htop-3.2.1-1.fc35_rename.x86_64.rpm'
Warning: Package 'wget-0:1.21.3-1.fc35.x86_64' has duplicate metadata entries, only one should exist
Warning:     Sourced from location: 'wget-1.21.3-1.fc35.x86_64.rpm'
Warning:     Sourced from location: 'foo/wget-1.21.3-1.fc35.x86_64.rpm'
Warning:     Sourced from location: 'bar/wget-1.21.3-1.fc35.x86_64.rpm'
Pool finished with errors
```

j-mracek:
The change looks logical but it will brake Copr and other workflows that previously worked (main reason why I marked it as blocked). The change does not describe explicitly why it is required but it is essential. I can only guess that it will provide a performance improvement but it is only my guess.

dralley:
@Conan-Kudo @j-mracek What if we leave the warnings but don't set the error code.

j-mracek:
I think that a warning will not brake anything and it contains interesting piece of information therefore I am fine with a warning. If there will be a complain we can easily turn the warning into info message.

src/createrepo_c.c
```diff
@@ -427,10 +427,10 @@ cr_create_repomd_records_for_groupfile_metadata(const cr_Metadatum *group_metada
 }
 
 /** Creates list of cr_RepomdRecords from list
- *  of additional metadata (cr_Metadatum) 
+ *  of additional metadata (cr_Metadatum)
  *
  * @param additional_metadata       List of cr_Metadatum
- * @param repomd_checksum_type      
+ * @param repomd_checksum_type
```
praiskup, line 433: Please avoid unnecessary white space changes.

praiskup:
Thank you for the work here, and for pinging me.   Yes, with non-zero exit status this would start failing in Copr for many repositories, unfortunately.  Warning would be completely OK, and "opt-in" non-zero exit status as well.  If we did this "opt-out", like "--ignore-duplicate-nvrs", we would have to coordinate with an appropriate fix in Copr upstream.

src/createrepo_c.c
praiskup: Just a nit, do we want to sort reverse-alphabetically?

src/dumper_thread.c
```diff
@@ -550,6 +550,19 @@ cr_dumper_thread(gpointer data, gpointer user_data)
...
+    g_mutex_lock(&(udata->mutex_nevra_table));
+    gchar *nevra = cr_package_nevra(pkg);
+    GArray *pkg_locations = g_hash_table_lookup(udata->nevra_table, nevra);
+    if (!pkg_locations) {
+        pkg_locations = g_array_new(FALSE, TRUE, sizeof(gchar *));
+        g_hash_table_insert(udata->nevra_table, nevra, pkg_locations);
+    }
+    gchar *location = g_strdup(pkg->location_href);
+    g_array_append_val(pkg_locations, location);
+    g_mutex_unlock(&(udata->mutex_nevra_table));
```
praiskup, line 564: It looks pretty innocent, (O(logN) I hope.  But actually we might need to opt out even the calculating of this hash.  In Copr with the [giant repositories like this](https://download.copr.fedorainfracloud.org/results/@rubygems/rubygems/fedora-rawhide-x86_64/repodata/), we use ` --no-stat --update --recycle-pkglist` to speedup metadata updates, which in turn means that createrepo_c isn't doing fancy with the `pkg` structure.  Now we use it to calculate the hash.

This deserves some benchmark IMO.

dralley: I'm running this command currently with the nonmodified `createrepo_c` and it is taking upwards of 10 minutes, is that expected?

dralley: Yeah, so, this is why.  85% of the runtime is spent inside of `g_queue_insert_sorted()`  That is crazy.  Given this, I'm not sure measuring any additional impact of my code is a great use of time. I don't expect it to even register, given that producing and compressing multiple gigabytes of metadata just barely registers as a tiny blip way over on the right-hand side of the chart.

flamegraph

dralley:
I have removed the error status and fixed the accidental formatting changes.

src/dumper_thread.c
dralley: Filed as https://github.com/rpm-software-management/createrepo_c/issues/323

praiskup: Nice analysis, @dralley!

praiskup:
In Copr, we are unable to (or we don't want to) control how our users take care of NVRs. Maybe I'm missing something, do you mean the opt-out/opt-in option?  Feel free to fill the [issue here](https://pagure.io/copr/copr/issues).

src/createrepo_c.c
```diff
@@ -1306,9 +1308,51 @@ main(int argc, char **argv)
...
+    gpointer key, value;
+
+    g_hash_table_iter_init(&iter, user_data.nevra_table);
+    while (g_hash_table_iter_next(&iter, &key, &value))
+    {
+        gchar *nevra = (gchar *) key;
+        GArray *locations = (GArray *) value;
+        if (locations->len > 1) {
+            g_warning("Package '%s' has duplicate metadata entries, only one should exist", nevra);
+            int strsort(const gpointer a, const gpointer b) {
```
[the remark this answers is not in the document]
kontura: You are correct, for example it doesn't compile with `clang`.
Please define it outside as a `static` function.

praiskup:
I rather meant that Copr allows the users to build whatever NVR they want, even multiple times, without additional constraints like "fail the build when the NVR already exists".  Copr does the build, places the results into an appropriate directory, and runs createrepo.  If createrepo fails, the build fails as well.

praiskup:
*Technically* it is possible (both the build ID and NVR is known and available), but automatic removals are not desired.  All builds are preserved for [at least 14 days](https://docs.pagure.org/copr.copr/user_documentation.html#how-long-do-you-keep-the-builds) - during this period it is up to the project maintainer to remove manually what they prefer.

kontura (approved):

dralley:
@praiskup Still, this is what confuses me - the shadowed releases would be uninstallable, I don't even know that you could "downgrade" to them as that might require a different NEVRA (not sure about this).

Thus whether they're kept or not is not a very consequential decision, and the choice to keep them might be a false choice as they're not really available in a true sense, unless you go hunt the packages down manually.

dralley:
Sidenote, once the new version makes it to the repos it would be neat to see if the performance fix makes a meaningful difference to the overall COPR performance metrics, if that stuff gets measured.

kontura:
> @praiskup Still, this is what confuses me - the shadowed releases would be uninstallable, I don't even know that you could "downgrade" to them as that might require a different NEVRA (not sure about this).

You can't, at least not with `dnf downgrade` command. I tried it now and it seems that the only way to get the "other" rpm with dnf would be `reinstall` with path directly to it which bypasses the repo anyway. 

> 
> Thus whether they're kept or not is not a very consequential decision, and the choice to keep them might be a false choice as they're not really available in a true sense, unless you go hunt the packages down manually.

I am not sure if this is what you mean and I am just repeating it but perhaps the older builds could be still kept in COPR but they would be excluded from the repo. From user perspective almost nothing would change (reposync wouldn't download the other pkgs anymore?).

dralley:
Right, and that feels like a viable option

>I am not sure if this is what you mean and I am just repeating it but perhaps the older builds could be still kept in COPR but they would be excluded from the repo.

Yup that's exactly what I meant.

praiskup:
Yes, user can manually decide what build in particular (from the duplicated builds) is to be deleted.  The delete action then triggers a build removal, and an automatic createrepo run.  So while the RPMs are not installable all the time they exist, it is just a non-problem situation, and it can become installable again.  This is just a result of some user/scripting error, easily fixable by the user.

Ignoring the directories that provide duplicate RPMs is technically possible, but potentially a very expensive feature (both for computational power and for the eventual code contribution).  And it would bring a zero gain to anyone.

Nb. this would be a serious backward incompatibility anyway, this could twist the createrepo users (not just Copr) into an inconvenient situation.  If the exit status is ever to be switched, we should have some opt-out option to absorb the change.

PR #324, opened 2022-06-25
Fix bad performance with task queue management

### LINKED ISSUE #323: Severe performance issue with large repositories
With large repositories, `createrepo_c` spends far more time sorting the queue of tasks than it does performing useful work.

With the repository https://download.copr.fedorainfracloud.org/results/@rubygems/rubygems/fedora-rawhide-x86_64/repodata/

Performing the command `createrepo_c --skip-stat --update --recycle-pkglist  .`

Produces the following flamegraph:

flamegraph

You can see that 85% of the time is spent inside of `g_queue_insert_sorted()`, of which about half is spent performing string comparisons via the sort operator.  The goal is to add new tasks to the threadpool in a sorted order based on the name of the file they operate on - but because this is a queue backed by a linked-list, and this repo contains more than 270,000 packages, this requires an enormous number of comparisons and pointer traversals, leading to extreme inefficiency.

  https://github.com/rpm-software-management/createrepo_c/blob/ace4c87a392b8c25fd7127da49516dba52d397c9/src/createrepo_c.c#L112=

Theoretically you might expect that if the tasks are already in a roughly-sorted order prior to being added to the queue, there might be magnification based on the fact that every new task would need to be inserted near the end.  Thus, roughly O(N^2)

dralley:
closes #323


dralley:
I don't really understand why the test is failing, it seems like it works fine?

dralley:
I can't tell if that helped.  The test process doesn't report any failures but it still returned exit code 2

kontura:
First of all thank you @dralley for all the work and help!

> I don't really understand why the test is failing, it seems like it works fine? I might be misunderstanding the test. Is there a way to run the test suite locally

You can checkout the output in the [action](https://github.com/rpm-software-management/createrepo_c/runs/7065637321?check_suite_focus=true), it is quite long but there is a bunch of fails because createrepo_c segfaults.

It can be run locally but there are a few additional steps involved, you can check it out here: https://github.com/rpm-software-management/ci-dnf-stack/ - the README has some instructions. The createrepo_c test suite can also be run without containers but that requires a bit more fiddling.

> 
> edit: Great, the number of failures varies from run to run

That is unrelated, there are a couple of zchunk tests that fail from time to time because zchunk segfaults. I have it on my ever elongating TODO list. Sorry about that.

---

I have investigated and the problem is that it segfaults in `task_cmp()`. Apparently the [CompareFunc](https://docs.gtk.org/glib/callback.CompareFunc.html) is called with different arguments for `g_queue_insert_sorted()` vs. `g_array_sort_with_data()` (`g_array_sort()`). 

While `g_queue_insert_sorted()` calls it with a pointer to the data, `g_array_sort()` calls it with a pointer to pointer to data. Like in [this example](https://docs.gtk.org/glib/type_func.PtrArray.sort.html).

So in the `task_cmp()` it needs something like:
```c
const struct PoolTask *a = *(struct PoolTask **) a_p;
const struct PoolTask *b = *(struct PoolTask **) b_p;
```

dralley:
Ironically, I already knew this from the previous change, but didn't think about it here.  <sigh>

I wonder why it didn't just crash immediately when I was using it.

Issue #306, opened 2022-02-16
createrepo_c silently parses bad metadata if the repository contains the same package (same pkgid, same NEVRA) multiple times


dralley:
The parsing API used by createrepo_c (the examples at least) is prone to parsing incorrect metadata in some scenarios where the metadata being parsed is, itself incorrect.  But in this case the parsed metadata is incorrect in a different way.

If `pkgcb` inserts a package object into a dictionary / hashmap type keyed by the `pkgid`, then the second occurrence of duplicate pkgid replaces the initial one.

But *after* that happens, `newpkgcb` is used to with the other metadata files, and it gets the package from the dictionary / hashmap and adds files and changelog metadata to it.  The end result is that the dictionary / hashmap will contain only one of the original two packages listed in the metadata, but that package will contain extra copies of every file and changelog.

As described [here](https://github.com/rpm-software-management/createrepo_c/issues/305#issuecomment-1041545459) this would also likely be true even with the `parse_main_metadata_together()` API.

This issue is compounded because `createrepo_c` doesn't prevent such incorrect repositories from being created: https://github.com/rpm-software-management/createrepo_c/issues/307

PR #325, opened 2022-08-10
Add --duplicated-nevra "keep-last" option, and --delayed-dump

praiskup:
This is follow-up for #307.  Unfortunately, the parallel processing performance degrades a bit
with the "keep-last" option.

Motivation for Copr: https://pagure.io/copr/copr/issue/840

We can tweak this in the future (keep latest pkg per mtime, keep the first one, etc.).

### DIFF
--- a/src/cmd_parser.c
+++ b/src/cmd_parser.c
@@ -67,8 +67,27 @@ struct CmdOptions _cmd_options = {
         .recycle_pkglist            = FALSE,
 
         .keep_all_metadata          = TRUE,
+        .nevra_duplicates           = CR_ARG_DUP_NEVRA_KEEP_ALL,
     };
 
+
+gboolean
+duplicated_nevra_option_parser(const gchar *,
+                               const gchar *value,
+                               gpointer,
+                               GError **error)
+{
+    if (!g_strcmp0(value, "keep"))
+        _cmd_options.nevra_duplicates = CR_ARG_DUP_NEVRA_KEEP_ALL;
+    else if (!g_strcmp0(value, "keep-last"))
+        _cmd_options.nevra_duplicates = CR_ARG_DUP_NEVRA_KEEP_LAST;
+    else {
+        g_set_error(error, ERR_DOMAIN, CRE_BADARG,
+                    "Bad --duplicated-nevra argument, use 'keep' or 'keep-last'.");
+        return FALSE;
+    }
+    return TRUE;
+}
 
 
 // Command line params
@@ -213,6 +232,8 @@ static GOptionEntry cmd_entries[] =
       "Read the list of packages from old metadata directory and re-use it.  This "
       "option is only useful with --update (complements --pkglist and friends).",
       NULL },
+    { "duplicated-nevra", 0, 0, G_OPTION_ARG_CALLBACK, duplicated_nevra_option_parser,
+      "What to do about duplicates.", NULL, },
     { NULL, 0, 0, G_OPTION_ARG_NONE, NULL, NULL, NULL },
 };
 
--- a/src/createrepo_c.c
+++ b/src/createrepo_c.c
@@ -578,6 +578,67 @@ static int strlensort(gconstpointer a, gconstpointer b)
     }
 }
 
+// Sorting function for DuplicateLocation pointers, by pkg build time.
+// Compatible with g_array_sort()
+static int buildtimesort(gconstpointer a, gconstpointer b)
+{
+    // Function is supposed to take a double-pointer so unfortunately you cannot pass a
+    // string-comparison function directly.
+    struct DuplicateLocation *a_loc = (struct DuplicateLocation *)a;
+    struct DuplicateLocation *b_loc = (struct DuplicateLocation *)b;
+
+    assert(a_loc->pkg->time_build != 0);
+    assert(b_loc->pkg->time_build != 0);
+
+    // order by build time first
+    int64_t result = a_loc->pkg->time_build - b_loc->pkg->time_build;
+    if (result)
+        return result;
+
+    // and then alphabetically by the rpm location
+    return g_strcmp0(a_loc->location, b_loc->location);
+}
+
+
+static int
+handle_nevra_duplicates(GArray *locations, CmdDupNevra option)
+{
+    int skipped = 0;
+    for (size_t i=0; i<locations->len; i++) {
+        struct DuplicateLocation location = g_array_index(
+                locations, struct DuplicateLocation, i);
+        if (option == CR_ARG_DUP_NEVRA_KEEP_LAST) {
+            if (i < locations->len - 1) {
+                location.pkg->skip_dump = TRUE;
+                skipped += 1;
+            }
+        }
+    }
+    return skipped;
+}
+
+
+static void
+duplicates_warning(const char *nevra, GArray *locations, CmdDupNevra option)
+{
+  g_warning("Package '%s' has duplicate metadata entries, only one should exist", nevra);
+
+  char *skip_reason= "";
+  if (option == CR_ARG_DUP_NEVRA_KEEP_LAST) {
+      skip_reason = " (not dumped, 'keep-last')";
+  }
+  for (size_t i=0; i<locations->len; i++) {
+      struct DuplicateLocation location = g_array_index(locations, struct
+                                                        DuplicateLocation, i);
+      g_warning("    Sourced from location: \'%s\', build timestamp: %ld%s",
+                location.location,
+                location.pkg->time_build,
+                location.pkg->skip_dump ? skip_reason : "");
+
+  }
+}
+
+
 int
 main(int argc, char **argv)
 {
@@ -734,6 +795,7 @@ main(int argc, char **argv)
     g_debug("Thread pool ready");
 
     long task_count = 0;
+    long package_count_in_headers = 0;
     GSList *current_pkglist = NULL;
     /* ^^^ List with basenames of files which will be processed */
 
@@ -778,6 +840,11 @@ main(int argc, char **argv)
 
     g_debug("Package count: %ld", task_count);
     g_message("Directory walk done - %ld packages", task_count);
+
+    if (cmd_options->nevra_duplicates)
+        // we need to construct the large table of cr_Packages to analyse
+        // all the NEVRAs together.
+        cmd_options->delayed_dump = TRUE;
 
     user_data.task_count        = task_count;
     if (cmd_options->delayed_dump)
@@ -1067,6 +1134,7 @@ main(int argc, char **argv)
         cr_xmlfile_set_num_of_pkgs(pri_cr_file, task_count, NULL);
         cr_xmlfile_set_num_of_pkgs(fil_cr_file, task_count, NULL);
         cr_xmlfile_set_num_of_pkgs(oth_cr_file, task_count, NULL);
+        package_count_in_headers = task_count;
     }
 
     // Open sqlite databases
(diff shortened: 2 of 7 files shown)


praiskup:
> parallel processing performance degrades a bit

By that I mean:
- using `createrepo_c --no-database --update --skip-stat --recycle-pkglist --pkglist add .  --workers 8`
- on a repository with 250k+ packages, and only one is being added, i.e. mostly a CPU bound task (not I/O)
- the slow-down is from 55s to 91s on my machine

We could start another pool of workers there, but I'm not sure it is needed at this point in time.

dralley:
Presumably in a situation where there are many duplicates, this would significantly improve the metadata size correct?

praiskup:
> Presumably in a situation where there are many duplicates, this would
significantly improve the metadata size correct?

Yes.  The delayed-write logic can also be tweaked to implement any
additional filtering mechanism in the future (if metadata size is a
concern).

dralley:
`keep-last` should probably keep the most recent package as determined by a second-order sort by build time, which is what yum / dnf / zypper does when encountering multiple packages of the same NEVRA.

kontura (changes requested):
> `keep-last` should probably keep the most recent package as determined by a second-order sort by build time, which is what yum / dnf / zypper does when encountering multiple packages of the same NEVRA.

I would also like the build-time ordering. It would break some current CI tests but I can easily fixed that.

src/createrepo_c.c
```diff
@@ -1416,7 +1462,7 @@ main(int argc, char **argv)
      * that unfortunately means we have to decompress metadata files change package
      * count value and compress them again.
      */
-    if (user_data.package_count != user_data.task_count){
+    if (user_data.package_count != user_data.task_count - skipped_pkgs){
```
kontura, line 1465: The `user_data.task_count` is also used in the [`cr_rewrite_pkg_count_thread()`](https://github.com/rpm-software-management/createrepo_c/blob/master/src/threads.c#L134) which causes problems because the `skipped_pkgs` are not taken into account. 

It can be reproduced by running with invalid packages (such as `echo "test" > bad.rpm`).

praiskup:
Thank you for the review!  Updated but WIP, I still need to implement build-time ordering.

src/dumper_thread.c
```diff
@@ -233,6 +233,54 @@ write_pkg(long id,
...
+void
+cr_delayed_dump_run(gpointer user_data)
+{
+    GError *tmp_err = NULL;
+    struct UserData *udata = (struct UserData *) user_data;
+    long int stop = udata->task_count;
+    g_debug("Performing the delayed metadata dump");
+    for (int id = 0; id < stop; id++) {
+        struct DelayedTask dtask = g_array_index(udata->delayed_write,
+                                                 struct DelayedTask, id);
```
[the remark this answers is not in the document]
praiskup: Good point, all IDs are now properly initialized in the delayed list.

praiskup:
> keep-last should probably keep the most recent package as determined by a second-order sort by build time, which is what yum / dnf / zypper does when encountering multiple packages of the same NEVRA.

So, trying to understand what actually DNF does, and it appears it prefers the oldest package from all with the NVR, aka the oldest per build time.  Is it a correct observation?

dralley:
AFAIK it prefers the newest per NEVRA, and the newest per build-time between packages with the same NEVRA.

kontura:
For dnf I think this depends on the order of packages in metadata. I believe libsolv picks just the **first** package in the metadata.
I made a hacky CI test to demonstrate: https://github.com/rpm-software-management/ci-dnf-stack/pull/1173

However this is just for `dnf install package-name` we might be doing something else for different arguments/commands. 
It is possible that different filtering will send different `id`s to libsolv resulting in installing the other package but I don't think there is any dedicated logic to it. It might happen by accident if at all.

praiskup:
Still experimenting a bit, and DNF seems to prefer the one which comes first
sorted alphabetically (or that comes first in the repodata, because createrepo_c
sorts the packages alphabetically).

For example [this repo](https://copr-be.cloud.fedoraproject.org/archive/issues/createrepo_c-pr-325/repo2/), in a Rawhide container:
```
# dnf -y install dummy-pkg --repofrompath=ordering,https://copr-be.cloud.fedoraproject.org/archive/issues/createrepo_c-pr-325/repo1/ --setopt ordering.gpgcheck=0
...
Installed:
  dummy-pkg-20221031_1057-1.fc38.x86_64 
...
# tail /var/log/dnf.librepo.log | grep dummy-pkg
https://copr-be.cloud.fedoraproject.org/archive/issues/createrepo_c-pr-325/repo2/a/dummy-pkg-20221031_1057-1.fc38.x86_64.rpm
```

The one starting with `../a/dummy-pkg...` was installed.  But when you look at RPM headers:

```
# for rpm in `find -name '*x86_64.rpm'`; do echo "== $rpm =="; rpm -qpi "$rpm" 2>/dev/null| grep Build\ Date; done
== ./a/dummy-pkg-20221031_1057-1.fc38.x86_64.rpm ==
Build Date  : Mon 31 Oct 2022 10:00:35 AM UTC
== ./b/dummy-pkg-20221031_1057-1.fc38.x86_64.rpm ==
Build Date  : Mon 31 Oct 2022 10:00:35 AM UTC
== ./c/dummy-pkg-20221031_1057-1.fc38.x86_64.rpm ==
Build Date  : Tue 01 Nov 2022 08:59:25 AM UTC
== ./d/dummy-pkg-20221031_1057-1.fc38.x86_64.rpm ==
Build Date  : Mon 31 Oct 2022 10:00:21 AM UTC
```

Then the newest one is `./c` which was built today.  Though I agree that what
you describe @dralley is the behavior I'd expect and prefer.  @j-mracek is this a bug,
accident, or expected behavior? :-)

praiskup:
@kontura was faster! :-)  Good to have this tested.

Anyway, for the `keep-last` option; should we do a two-level sort, by build time (which might still have collisions) and then by path name alphabetically?

kontura:
> Anyway, for the `keep-last` option; should we do a two-level sort, by build time (which might still have collisions) and then by path name alphabetically?
 
That sounds reasonable to me. :slightly_smiling_face:

praiskup:
Ok, please take another look.

src/createrepo_c.c
```diff
@@ -1340,30 +1415,45 @@ main(int argc, char **argv)
...
     gpointer key, value;
 
+    int skipped_pkgs = 0;
     g_hash_table_iter_init(&iter, user_data.nevra_table);
     while (g_hash_table_iter_next(&iter, &key, &value))
     {
         gchar *nevra = (gchar *) key;
         GArray *locations = (GArray *) value;
         if (locations->len > 1) {
-            g_warning("Package '%s' has duplicate metadata entries, only one should exist", nevra);
```
dralley, line 1349: We should keep the warning in place, A) because not all users will use the option, and B) because even users who do use the option should probably be aware of potential issues even if we're applying our own heuristics to deal with it.

dralley:
Are we sure libsolv has any deliberate strategy here at all, as opposed to some kind of parsing bug where various package entries attempt to overwrite each other?

Is there an easy way that we can ask libsolv "load this repo and then dump the solvables" to compare the input and what libsolv sees?

e.g. If some package was present twice in the metadata, would you see only one solvable, and would potentially have doubled up filelist and changelog metadata

[This comment makes me suspicious that this will be the case](https://github.com/openSUSE/libsolv/blob/master/ext/repo_rpmmd.c#L654-L668)

kontura:
> Are we sure libsolv has any deliberate strategy here, as opposed to some kind of parsing bug where various package entries attempt to overwrite each other?

Oh I meant during solving. Libsolv does parse all of the packages and each has a unique id. If you do `install foo` libdnf basically finds all ids that match name `foo` and sends them to libsolv. Here the solver has to pick something, I would guess that from duplicate packages it takes the lowest id (that is the first duplicate in metadata).

> 
> Is there an easy way that we can ask libsolv "load this repo and then dump the solvables" to compare the input and what libsolv sees at the pool level (pre- any kind of actual dependency solving / choosing logic)?

There is exactly such a thing. Libsolv has these [tools](https://github.com/openSUSE/libsolv/tree/master/tools) (rpm libsolv-tools-0.7.22-3.fc37.x86_64). You can do:
```
$ repo2solv ./path/to/repo | dumpsolv
```

src/dumper_thread.c
```diff
@@ -43,7 +43,7 @@ struct BufferedTask {
     long id;                        // ID of the task
     struct cr_XmlStruct res;        // XML for primary, filelists and other
     cr_Package *pkg;                // Package structure
-    int pkg_from_md;                // If true - package structure if from
+    gboolean clean;                 // If false - package structure if from
```
kontura, line 46: This comment now has twice `If false..`. I suppose the second should be changed to `If true..`?

Also on the first line there is `If false - package structure if from` I think that should have been `If false - package structure is from`. It would be nice if we could fix that as well.

dralley:
>Oh I meant during solving. Libsolv does parse all of the packages and each has a unique id (even duplicates).

I think that would work while parsing primary.xml since you can create new solvables, I'm just curious what happens if you encounter duplicate pkgid while parsing other.xml and filelists.xml.  The files and changelogs have to be associated back to solvables, and all it has to go off of is NEVRA and pkgid, so how would it pick which solvable the files and changelogs get associated with?  If they all get dumped onto one solvable, then you'd have duplicate solvables but with broken file deps and changelogs on some or all of them.

Basically similar to https://github.com/rpm-software-management/createrepo_c/issues/306

However I've just done some testing and nothing stood out as being obviously wrong, so maybe it has some magic up its sleeve.

src/createrepo_c.c
```diff
@@ -578,6 +578,66 @@ static int strlensort(gconstpointer a, gconstpointer b)
...
+    // and then alphabetically by the rpm location
+    return g_strcmp0(a_loc->location, b_loc->location);
+}
+
+
+static int
+handle_nevra_duplicates(GArray *locations, CmdDupNevra option)
+{
+    int skipped = 0;
+    for (int i=0; i<locations->len; i++) {
```
kontura, line 607: Here we compare signed `int i` with unsigned `locations->len`,
I know its already present elsewhere but these things show up in covscans and other analyzers.
Can you change the `i` to unsigned? To `size_t` perhaps?

src/createrepo_c.c
```diff
@@ -578,6 +578,66 @@ static int strlensort(gconstpointer a, gconstpointer b)
...
+  g_warning("Package '%s' has duplicate metadata entries, only one should exist", nevra);
+
+  char *skip_reason= "";
+  if (option == CR_ARG_DUP_NEVRA_KEEP_LAST) {
+      skip_reason = " (not dumped, 'keep-last')";
+  }
+  for (int i=0; i<locations->len; i++) {
+      struct DuplicateLocation location = g_array_index(locations, struct
+                                                        DuplicateLocation, i);
+      g_warning("    Sourced from location: \'%s\', build timestamp: %d%s",
```
kontura, line 632: The `time_build` is `gint64` so I think the format specified should be `%ld`.

src/cmd_parser.c
```diff
@@ -213,6 +213,11 @@ static GOptionEntry cmd_entries[] =
       "Read the list of packages from old metadata directory and re-use it.  This "
       "option is only useful with --update (complements --pkglist and friends).",
       NULL },
+    { "delayed-dump", 0, G_OPTION_FLAG_HIDDEN, G_OPTION_ARG_NONE,
```
kontura, line 235: Is there any benefit to having this as a hidden option? Maybe for debugging? But personally I wouldn't add it here at all.

praiskup: I used it several times for debugging, yes.  There shouldn't be cons in having the option, so I'd prefer to keeping it?  Just please re-consider, and I'll drop if you really want.

kontura:
> Basically similar to #306
> 
> However I've just done some testing and nothing stood out as being obviously wrong, so maybe it has some magic up its sleeve.

I was interested in how it is done so I spent some time digging into the libsolv parser and there is no magic. It has the same bug as you mentioned.
Its just that I didn't realize that `repo2solv` by default uses just primary. Testing it with options to use full filelists and changelogs:
```
repo2solv . -F -C | dumpsolv
```
It does add all the duplicate entries to the first `id`, duplicating them.

I am thinking what does this mean for dnf or other tools. dnf would show duplicated changelogs but the filelists probably just add some overhead?

praiskup:
PR updated.

src/cmd_parser.c
praiskup: ACK, dropped the command line option then, thanks for checking. I kept the "internal" delayed_dump option just to not overcomplicate the commit changes if that is OK?

kontura: Great, thanks!

kontura (approved):

dralley:
@kontura So Pulp has some logic to attempt to pick the package that DNF would pick when syncing a repo that happens to contain duplicate NEVRA (it skips over the other duplicates), and currently it uses build time as the differentiator because that's what we thought it does.

Should we change that to instead just pick the first one?

kontura:
I have tried syncing with DNF `reposync` and that keeps all the duplicates.
Just to make sure I understand correctly you mean when Pulp is doing the syncing right?

If they have duplicate NEVRA and pkgid (checksum) then libsolv picks the first one.

However for just duplicate NEVRA DNF works with all of them. For example when you try to install something both of them provide then yes it (libsolv) picks the first one. But if you do `install /path/to/file` and the file is only present in the second one - it uses the second one. 
I think that in general libsolv will use any package of the duplicates that satisfies the dependency resolution and since the packages are different it could be any of them.

Based on that if you want to match what DNF does you need all of them.

dralley:
> Just to make sure I understand correctly you mean when Pulp is doing the syncing right?

Yeah.  The trouble is that Pulp encoded some optimistic assumptions early on that NEVRA would be unique, and it led to bugs when that wasn't the case, but we can't reject those repos because there's too many of them and it would be a pain in the ass, so instead we're just trying to pick the one that DNF / Yum / zypper would prefer.

In theory the dependencies ought to be the same between the packages unless the packager has been doing grievously stupid things anyway.

PR #326, opened 2022-08-30
Keep location_base on update if no baseurl defined in args

achamo:
When using --update, --skip-stat and --reycle-pkglist we do not want to rewrite location_base but keep all the metadata for a package
By default createrepo_c removes the location_base that was already set in the metadata using the null value from the parameter
Skip it and keep the location_base when it's null (and if both are null it shouldn't change the current behaviour)

### DIFF
--- a/src/dumper_thread.c
+++ b/src/dumper_thread.c
@@ -631,10 +631,12 @@ cr_dumper_thread(gpointer data, gpointer user_data)
                 md->loadingflags &= ~CR_PACKAGE_SINGLE_CHUNK;
 
                 // We have usable old data, but we have to set proper locations
-                // WARNING! This two lines destructively modifies content of
-                // packages in old metadata.
+                // WARNING! location_href is overidden
+                // location_base is kept by default, unless specified differently
+                //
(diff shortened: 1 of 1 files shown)


dralley:
Do you think putting this:

```C
} else {
    location_base = pkg->location_base;
}
```

on line 430 might be more appropriate?

src/dumper_thread.c
```diff
@@ -476,7 +476,8 @@ cr_dumper_thread(gpointer data, gpointer user_data)
                 // WARNING! This two lines destructively modifies content of
                 // packages in old metadata.
```
dralley, line 477: The comment could use updating

kontura (changes requested):
Hello, sorry for the delay.
The change looks good, it does affect all runs with `--update` not just with `--skip-stat` but I think that is fine.

I have tried to resolve the conflicts but it created a new commit and its not ideal. 
If you are still interested can you rebase this?

achamo:
Thanks for the update, I'm still interested as I'm using it to generate my repositories! I would prefer to see this patch merged than keep maintaining my own branch.
I updated the comment in the code, as suggested!

kontura:
Thanks!

PR #328, opened 2022-10-19
Include <rpm/rpmstring.h> for rasprintf

thesamesam:
Fixes build with Clang 16.

Bug: https://bugs.gentoo.org/875698
Signed-off-by: Sam James <sam@gentoo.org>


kontura:
Thanks!

PR #334, opened 2023-01-06
Avoid delta regeneration if drpm exists and target_md5 matches

davidvoit:
If a drpm file for the same target already exists, we don't need to process the original files again.

We can't check the source rpm only the target rpm hashsum, but as drpm is content based not rpm based this should be could enough.

If the target is overridden with different content but has the same name, we need to recalculate the delta rpm.

### DIFF
--- a/src/deltarpms.c
+++ b/src/deltarpms.c
@@ -35,7 +35,8 @@
 #include "misc.h"
 #include "error.h"
 
-
+// MD5 support, private to work even without legacy support enabled
+#define CR_CHECKSUM_MD5INT 250
 #define ERR_DOMAIN      CREATEREPO_C_ERROR
 
 gboolean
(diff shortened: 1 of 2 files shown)


src/deltarpms.c
```diff
@@ -49,6 +50,43 @@ cr_drpm_support(void)
 
 #ifdef    CR_DELTA_RPM_SUPPORT
 
+gboolean
+cr_drpm_check_checksum(const char *drpmpath, const char *new_path) {
```
kontura, line 54: I think `target_md5` and `delta` could leak in the error cases.

kontura (changes requested):
Hello, I don't really like sneaking the `MD5` hash back in like this.
I think we should either enable this only when the `LEGACY_HASHES` are allowed or find a different way.

davidvoit:
MD5 drpms Always are using md5 sums even iegacy is disabled. So you only
break the caching then but producing drpm still works this seems odd to me

Maybe drpm should be updated If we really want to phase out md5

amatej ***@***.***> schrieb am Do., 9. Feb. 2023, 09:32:

kontura:
> Maybe drpm should be updated If we really want to phase out md5

Yes, if possible that would be the best option.

Also if this functionality was directly in drpm createrepo_c wouldn't need to anything about the used hash.

kontura:
I am closing this PR, since this is not the right place to add it.

If you are still interested in the feature please add it to [drpm](https://github.com/rpm-software-management/drpm) (probably in `drpm_make`(...)).

PR #335, opened 2023-01-12
Add an option to compile against zlib-ng instead of zlib

dralley:
Zlib-ng has roughly (ymmv) twice the performance of standard zlib.


dralley:
@praiskup I'm getting these compile errors and I really do not understand why, `zlib-ng.h` is definitely present on my system and it definitely declares those functions and types.  But I'm not a C dev so I may be missing something obvious.

Screenshot from 2023-01-12 01-00-02

CMakeLists.txt
```diff
@@ -81,6 +79,19 @@ IF (ENABLE_DRPM)
     ADD_DEFINITIONS("-DCR_DELTA_RPM_SUPPORT")
 ENDIF (ENABLE_DRPM)
 
+# Zlib-ng is a performance-oriented fork of zlib
+# with a more active development cadence
+OPTION (WITH_ZLIBNG "Build against zlib-ng rather than zlib" OFF)
+IF (WITH_ZLIBNG)
+    pkg_check_modules(ZLIB-NG REQUIRED)
```
dralley, line 86: Not sure if it's related but while this seems to work fine, none of the variables (like `ZLIBNG_INCLUDE_DIRS`, `ZLIBNG_LIBRARY`) get populated.

src/CMakeLists.txt
```diff
@@ -77,10 +77,15 @@ TARGET_LINK_LIBRARIES(libcreaterepo_c ${LZMA_LIBRARIES})
 TARGET_LINK_LIBRARIES(libcreaterepo_c ${OPENSSL_LIBRARIES})
 TARGET_LINK_LIBRARIES(libcreaterepo_c ${RPM_LIBRARIES})
 TARGET_LINK_LIBRARIES(libcreaterepo_c ${SQLITE3_LIBRARIES})
-TARGET_LINK_LIBRARIES(libcreaterepo_c ${ZLIB_LIBRARY})
 TARGET_LINK_LIBRARIES(libcreaterepo_c ${ZCK_LIBRARIES})
 TARGET_LINK_LIBRARIES(libcreaterepo_c ${DRPM_LIBRARIES})
 
+IF (WITH_ZLIBNG)
+  TARGET_LINK_LIBRARIES(libcreaterepo_c ${ZLIBNG_LIBRARY})
```
dralley, line 84: Variable not populated

dralley:
@praiskup I used that rubygem repo from COPR with 270,000 packages to test.  

The main issue is that without `--skip-stat --recycle-packagelist --update` the difference gets swamped by stat calls and reading headers, and without `--no-database` it gets swamped by the cost of BZ2 compression, which seems to be very inefficient in comparison to even zlib.

In the best case scenario using all 4 flags together, it does reduce time spent inside the compression library by about 40%, but that only equates to a 10% overall improvement in runtime, which might not be worthwhile relative to the maintenance burden.  zstd should be faster and result in a 10-20% improvement which is easier to justify.

But as far as COPR goes it would be a great idea to disable generating sqlite metadata everywhere except EL6 and EL7 repos.  It dropped the runtime from 242 seconds to 106 seconds and is definitely the best bang-for-buck change. 

DNF doesn't use it so it's safe to drop, but I think "mdapi" does still use it currently, I don't know if that's a blocker for COPR the way it is for Fedora proper https://pagure.io/releng/issue/10745

dralley:
This is a profile of a `--skip-stat --recycle-packagelist --update` run on the aforementioned repo with original zlib and including sqlite metadata.  You can see that the main issue is bz2 compression, specifically.

std-zlib-with-sqlite

praiskup:
Thank you for another look at performance.  We use --skip-stat
--recycle-packagelist --update for sure, otherwise it would be entirely
unusable.  Disabling sqlite data will help a lot, and is [tracked here](https://github.com/fedora-copr/copr/issues/1171).
Going to zstd seems to be a long-term task (not sure how likely is to get
this to e.g. RHEL9+ at least).

dralley:
It *should* be supported by RHEL 8+ unless it's compiled out when building libsolv.  It's hard to test without having a repo first.

PR #336, opened 2023-01-12
Disable hardware accelerated DEFLATE support on s390x for tests

dcantrell:
s390x has a hardware implementation of DEFLATE, which changes how zlib and gzip works.  For the purposes of the test suite, disable the hardware accelerated DEFLATE usage so that our gzipped test files will always generate the same message digests regardless of architecture.

= changelog =
msg: Set DFLTCC=0 on s390x for tests; disable hw DEFLATE support
type: bugfix
resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2130179

Signed-off-by: David Cantrell <dcantrell@redhat.com>

### DIFF
--- a/tests/run_tests.sh.in
+++ b/tests/run_tests.sh.in
@@ -9,6 +9,13 @@ RET=0
 # object file: No such file or directory" error message.
 export "LD_LIBRARY_PATH=${CMAKE_BINARY_DIR}/src/:"
 
+# s390x has a hardware implementation of DEFLATE, which changes how
+# zlib and gzip works.  For the purposes of the test suite, disable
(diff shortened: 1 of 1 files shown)


dcantrell:
Yes, it is that test.  And that solution will work too.  As long as the test is not assuming a specific SHA-256 value but rather looking at the decompressed contents, it should work fine with either software implemented DEFLATE or the hardware implemented DEFLATE on z.

dralley:
@dcantrell Addressed here, if you would like to provide a review https://github.com/rpm-software-management/createrepo_c/pull/339

dcantrell:
This PR should be closed in favor of PR #339

kontura:
ok, I am closing it.

PR #341, opened 2023-01-27
replace 'cp' binary execution with gio

Kangie:
While attempting to use createrepo_c in a lightweight container I discovered that the utility does not work with all cp implementations (i.e. Alpine's busybox, likely others)

This PR amends `cr_cp` to use GIO for file copy operations rather than calling a separate binary on the system (as a bonus, the resulting code is simpler)

`cr_rm` could probably use a similar fix if any developers are interested - it's basically a copy+paste of the cr_cp code to handle recursion and flags. It's less pressing - I'm not aware of an implementation of `rm` that doesn't support both `r` and `f` flags!

Of note is that we now only use the `G_FILE_COPY_ALL_METADATA` flag - the previous recursive flag was redundant, we either need a recursive copy (because the src is a directory) or we don't (because it's a file).


Kangie:
I'm not a C programmer (or _really_ a programmer tbh!)

Any feedback is welcome and will be addressed promptly - I already know that there will be some work required to get this passing CI.

kontura (changes requested):
Hello, thank you for the contribution.

This change currently breaks API so we cannot merge it as is.
I think there are basically two options:
1. Change `cr_cp(..)` only on the inside but keep accepting the same params.
or
2. Introduce a new function for copying, use the new one in createrepo_c and mark the old one as deprecated with a warning perhaps.

It seems to me the current signature of `cr_cp(..)` could work fine but it is missing the `GCancellable *cancellable` so the question is if it is worth it to add a new function because of it?

Kangie:
@kontura

I wasn't aware of your stability guarantees. The version numbering made me think this might be ok. Could you give me a reference?

My assumption was that you're using semver.

kontura:
I think we are basically using semver. 

The way I see it: if we do this we would have to increase the MAJOR version. Now we would like to do more changes in a new MAJOR version like fixing some of the `createrepo_c` CLI options and other stuff. The problem is we don't have resources to work on that now so your change wouldn't be released for possibly a long time.

If that is fine with you we can perhaps keep it open and add some label or create a new branch or something.

Kangie:
> I think we are basically using semver.
> 
> The way I see it: if we do this we would have to increase the MAJOR version. Now we would like to do more changes in a new MAJOR version like fixing some of the `createrepo_c` CLI options and other stuff. The problem is we don't have resources to work on that now so your change wouldn't be released for possibly a long time.

Hi @kontura. You're mostly correct in your understanding of semver, however you've missed the special case: Major version 0 has [no guarantees of API stability](https://semver.org/#spec-item-4).

At the end of the day I just want the software to work. I'm happy to write a `do_copy_properly` function and revert any changes to `cr_cp`, it just doesn't seem necessary at this point based on the version numbering and your use of semver; until you have a 1.0.0 release you can do whatever you want with your API.

There's probably a separate discussion about whether this software should be at least major version '1' - it's certainly available out in the wild!

If you have a strong preference for new function I'm happy to do the implementation; it'll be trivial at this point.

kontura:
oh, I see, I think we should treat `createrepo_c` as stable. You are correct the MAJOR version shouldn't be 0 for some time now. It's been too long and there are too many users.

In the end yes I would strongly prefer if we didn't break the API. :slightly_smiling_face:

Kangie:
I'm not sure why CI/CD failed - it seems to relate to packaging python?

eli-schwartz:
Spec file expects old version number in the egg-info file. :)

src/misc.c
```diff
@@ -1396,33 +1394,41 @@ cr_run_command(char **cmd, const char *working_dir, GError **err)
...
 gboolean
-cr_cp(const char *src,
-      const char *dst,
-      cr_CpFlags flags,
-      const char *working_dir,
+cr_gio_cp(GFile *src,
+      GFile *dst,
+      GFileCopyFlags flags,
+      GCancellable *cancellable,
       GError **err)
```
eli-schwartz, line 1401: This deletes the cr_cp function and replaces it with cr_gio_cp, but does not implement an additional copy of cr_cp (potentially as a thin wrapper over cr_gio_cp). However, misc.h only appended a new function, so the header says cr_cp exists but no function is ever actually compiled.

This seems to go counter to the above discussion... did you forget to re-add that in?

The update diff is https://github.com/rpm-software-management/createrepo_c/compare/7f99f9a675a58ba4cbf76e1a86b3c30301bcc13d..6a9a871abcfd2685a724f5ed1428d4de56ff5bb6 and shows an re-inclusion in misc.h but not misc.c

eli-schwartz: Also as discussed elsewhere, `__attribute((deprecated("please use cr_gio_cp instead")))` or something to that general effect would be apropos. :)

VERSION.cmake
```diff
@@ -1,3 +1,3 @@
-SET(CR_MAJOR "0")
-SET(CR_MINOR "20")
-SET(CR_PATCH "1")
+SET(CR_MAJOR "1")
```
kontura, line 1: We have discussed this in our team and we probably don't want to do this change just now even though we have stable API.
The change it self could confuse users especially since we don't do any breaking changes.
Can you please drop it? 
Otherwise it looks good to me.

kontura (changes requested):

VERSION.cmake
Kangie: I'd note that the change in semver from 0.x to 1.x **is** arbitrary; it's whenever you think your API is mature enough to derlare it stable. If you're using semver and (as you've claimed) have a stable API you **must** have a major version of at least `1`.

I'm happy to revert the change, but, why do you think that your software declaring that it's mature enough to be v1.0.0 would scare users?

As an end-user myself, I'm far more scared of relying on a critical piece of software that (based on its versioning) is in _perpetual_ alpha.

Kangie: another note - if you don't set the major version to one here (which won't have any impact until you tag that release with whatever other changes you decide to include) you're **not** actually using semver. You're using "rpm-software-management's almost-but-not-quite semver" which will just add to the confusion going forward.

eli-schwartz: Well no, it's using semver to declare a semver stability level of "whether we intend to break API or not, if we do end up doing so it's not a bug and you just have to deal with it".

:D

I do agree that a project committing to 1.0 is definitely not going to confuse users, ever.

eli-schwartz: Specifically, according to https://semver.org

> 4. Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.
> 
> 5. Version 1.0.0 defines the public API. The way in which the version number is incremented after this release is dependent on this public API and how it changes.

And:
> 8. Major version X (X.y.z | X > 0) MUST be incremented if any backwards incompatible changes are introduced to the public API. 

Notice that bumping the major version from 0 to 1 is rigidly defined as an exception to the normal rule that bumping the major version indicates backwards incompatible changes "to the public API".

Instead, bumping from 0 to 1 means that an unstable private API has now been marked as public. Rule 8 is only applicable to major versions >= 1

kontura: I think I agree with you, I will try to push for 1.0.0.

Thank you for the inputs and the PR!

Issue #337, opened 2023-01-13
Compile warnings


dralley:
```
/home/dalley/devel/community/createrepo_c/src/createrepo_c.c: In function ‘allowed_file’:
/home/dalley/devel/community/createrepo_c/src/createrepo_c.c:77:13: warning: ‘g_pattern_match’ is deprecated: Use 'g_pattern_spec_match' instead [-Wdeprecated-declarations]
   77 |             if (g_pattern_match((GPatternSpec *) element->data,
      |             ^~
In file included from /usr/include/glib-2.0/glib.h:65,
                 from /home/dalley/devel/community/createrepo_c/src/createrepo_c.c:20:
/usr/include/glib-2.0/glib/gpattern.h:50:15: note: declared here
   50 | gboolean      g_pattern_match          (GPatternSpec *pspec,
      |               ^~~~~~~~~~~~~~~

```

```
/home/dalley/devel/community/createrepo_c/src/checksum.c: In function ‘cr_checksum_type’:
/home/dalley/devel/community/createrepo_c/src/checksum.c:53:23: warning: writing 1 byte into a region of size 0 [-Wstringop-overflow=]
   53 |         name_lower[x] = tolower(name[x]);
      |                       ^
/home/dalley/devel/community/createrepo_c/src/checksum.c:43:10: note: at offset 8 into destination object ‘name_lower’ of size 8
   43 |     char name_lower[MAX_CHECKSUM_NAME_LEN+1] = {0};
      |          ^~~~~~~~~~

```

dralley:
`g_pattern_spec_match` and `g_pattern_match` were added and deprecated (respectively) in glibc 2.70, which is pretty new.  It's probably not worth fixing for now, as it would still need to use the old function when built on EL7 or EL8

PR #342, opened 2023-01-27
Use g_pattern_spec_match() with glib >= 2.70.0

dcantrell:
Starting with glib-2.70.0, the g_pattern_match() function has been deprecated in favor of g_pattern_spec_match().  Still support building with older versions of glib, just add some preprocessor handling to eliminate the build time deprecation warning.

msg: Use g_pattern_spec_match() with glib >= 2.70.0 type: enhancement

Signed-off-by: David Cantrell <dcantrell@redhat.com>

### DIFF
--- a/src/createrepo_c.c
+++ b/src/createrepo_c.c
@@ -59,6 +59,16 @@
 #define OUTDELTADIR "drpms/"
 #define DEFAULT_DATABASE_VERSION    10
 
+/*
+ * Starting with glib 2.70.0, g_pattern_spec_match() replaces
+ * g_pattern_match().
+ */
+#ifdef GLIB_VERSION_2_70
+#define PATTERN_MATCH g_pattern_spec_match
+#else
+#define PATTERN_MATCH g_pattern_match
+#endif
+
 /** Check if the filename is excluded by any exclude mask.
  * @param filename      Filename (basename).
  * @param exclude_masks List of exclude masks
@@ -74,8 +84,8 @@ allowed_file(const gchar *filename, GSList *exclude_masks)
 
         GSList *element = exclude_masks;
         for (; element; element=g_slist_next(element)) {
-            if (g_pattern_match((GPatternSpec *) element->data,
-                                str_len, filename, reversed_filename))
+            if (PATTERN_MATCH((GPatternSpec *) element->data,
+                              str_len, filename, reversed_filename))
             {
                 g_free(reversed_filename);
                 g_debug("Exclude masks hit - skipping: %s", filename);


kontura:
This resolves: https://github.com/rpm-software-management/createrepo_c/issues/337

Thanks!

andyone:
@dcantrell I have found that this commit breaks build on Oracle Linux 9 due to defined macro `GLIB_VERSION_2_70` for versions lower than 2.70:

```
$ rpm -q glib2-devel
glib2-devel-2.68.4-6.el9.x86_64

$ grep GLIB_VERSION_2_70 /usr/include/glib-2.0/glib/gversionmacros.h
 * GLIB_VERSION_2_70:
#define GLIB_VERSION_2_70       (G_ENCODE_VERSION (2, 70))
#if GLIB_VERSION_MIN_REQUIRED >= GLIB_VERSION_2_70
#if GLIB_VERSION_MAX_ALLOWED < GLIB_VERSION_2_70
```

```
/home/builder/rpmbuild/BUILD/createrepo_c-0.21.1/src/createrepo_c.c: In function 'allowed_file':
/home/builder/rpmbuild/BUILD/createrepo_c-0.21.1/src/createrepo_c.c:67:23: warning: implicit declaration of function 'g_pattern_spec_match'; did you mean 'g_pattern_match'? [-Wimplicit-function-declaration]
   67 | #define PATTERN_MATCH g_pattern_spec_match
      |                       ^~~~~~~~~~~~~~~~~~~~
/home/builder/rpmbuild/BUILD/createrepo_c-0.21.1/src/createrepo_c.c:87:17: note: in expansion of macro 'PATTERN_MATCH'
   87 |             if (PATTERN_MATCH((GPatternSpec *) element->data,
      |                 ^~~~~~~~~~~~~
…
[ 98%] Linking C executable createrepo_c
/usr/bin/ld: CMakeFiles/createrepo_c.dir/createrepo_c.c.o: in function `allowed_file':
/home/builder/rpmbuild/BUILD/createrepo_c-0.21.1/src/createrepo_c.c:87: undefined reference to `g_pattern_spec_match'
```

dralley:
@andyone That sounds like an Oracle Linux 9 problem, not a problem with the commit.  Why would they define `GLIB_VERSION_2_70` if they're using 2.68?  Maybe a backport mistake, but almost definitely wrong.

andyone:
I've made some research and found that all EL9 distros (OL, Alma, Rocky) have this problem, and you are right, it's a backport mistake.

Spec file contains [this lines](https://gitlab.com/redhat/centos-stream/rpms/glib2/-/blob/c9s/glib2.spec#L15-21):

```spec
# Add GPowerProfileMonitor
# https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1965
# https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2194
# https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2222
Patch: 1965.patch
Patch: 2194.patch
Patch: 2222.patch
```

`1965.patch` [appends](https://gitlab.com/redhat/centos-stream/rpms/glib2/-/blob/c9s/1965.patch) macro for 2.70.

dralley:
@dcantrell ^^^

j-mracek:
Thank you for verification that the issue is not on our side.

src/createrepo_c.c
```diff
@@ -59,6 +59,16 @@
 #define OUTDELTADIR "drpms/"
 #define DEFAULT_DATABASE_VERSION    10
 
+/*
+ * Starting with glib 2.70.0, g_pattern_spec_match() replaces
+ * g_pattern_match().
+ */
+#ifdef GLIB_VERSION_2_70
```
mcatanzaro, line 66: OK so I've been pointed to this issue. For future reference, a better way to do this would be `#if GLIB_CHECK_VERSION(2, 70, 0)`. The macro you used is intended for declaring API availability; for checking the GLib version in use, `GLIB_CHECK_VERSION` is the easiest way, but you can also use `GLIB_MAJOR_VERSION`, `GLIB_MINOR_VERSION`, and `GLIB_MICRO_VERSION`.

That said, I will change the RHEL 9 patch to drop use of `GLIB_VERSION_2_70` anyway since it's easy to do so. I had done it this way because I thought "we want to show the API as new in version 2.70 so that anybody using the RHEL 9 version of GLib doesn't think the API is present in upstream version 2.68," but it's possible to use `GLIB_VERSION_2_68` when declaring the API yet still show it as 2.70 in the documentation.

mcatanzaro: Actually I think I will leave things be, so that the GPowerProfileMonitor APIs remain treated as version 2.70 for the purposes of `GLIB_VERSION_MIN_REQUIRED` and `GLIB_VERSION_MAX_ALLOWED`. I think this should not be a problem for createrepo_c as long as you switch to using `GLIB_CHECK_VERSION`.

PR #343, opened 2023-01-30
Add support for libmodulemd on EL7

sharewax:
For building on EL7 with libmodulemd we need to use package libmodulemd2 from epel repo.


ppisar:
I confirm that upstream libmodulemd ≥ 2.3.0 is provided in EPEL 7 repository in libmodulemd2 RPM package. So this the pull request makes sense.

kontura:
Thank you both!

PR #345, opened 2023-01-31
Add conditional zstd compression support

kontura:
For: https://github.com/rpm-software-management/createrepo_c/issues/82

Including `zstd` is optional but defaults to `ON`.

I have converted PR: https://github.com/rpm-software-management/createrepo_c/pull/251 to not use `rpmio` but instead use `zstd` directly by basically inlining the rpm `zstd` functions.

I chose the compression level value `11` kind of arbitrarily, we may want to find a better value.


CMakeLists.txt
```diff
@@ -21,6 +21,8 @@ if(NOT BUILD_LIBCREATEREPO_C_SHARED)
   set(CMAKE_POSITION_INDEPENDENT_CODE 1)
 endif()
 
+option(WITH_ZSTD "Build with zstd support" ON)
```
dralley, line 24: Add this to the build deps section of the README

`* libzstd (<http://facebook.github.io/zstd/>) - libzstd-devel/libzstd-dev`

src/cmd_parser.c
```diff
@@ -177,7 +177,7 @@ static GOptionEntry cmd_entries[] =
     { "xz", 0, 0, G_OPTION_ARG_NONE, &(_cmd_options.xz_compression),
       "Use xz for repodata compression.", NULL },
     { "compress-type", 0, 0, G_OPTION_ARG_STRING, &(_cmd_options.compress_type),
-      "Which compression type to use.", "COMPRESSION_TYPE" },
+      "Which compression type to use. Supported compressions are: bzip2, gzip, zck, zstd, xz.", "COMPRESSION_TYPE" },
     { "general-compress-type", 0, 0, G_OPTION_ARG_STRING, &(_cmd_options.general_compress_type),
       "Which compression type to use (even for primary, filelists and other xml).",
```
dralley, line 182: Kind of unrelated change but, could you add clarification to `compress_type` and `general_compress_type`?

e.g.

```diff
     { "compress-type", 0, 0, G_OPTION_ARG_STRING, &(_cmd_options.compress_type),
-      "Which compression type to use.", "COMPRESSION_TYPE" },
+      "Which compression type to use for secondary metadata files (comps, updateinfo, etc). Supported compressions are: bzip2, gzip, zck, zstd, xz.", "COMPRESSION_TYPE" },
     { "general-compress-type", 0, 0, G_OPTION_ARG_STRING, &(_cmd_options.general_compress_type),
-      "Which compression type to use (even for primary, filelists and other xml).",
+      "Which compression type to use (even for primary, filelists and other xml). Supported compressions are: bzip2, gzip, zck, zstd, xz.",
       "COMPRESSION_TYPE" },

```

createrepo_c.spec
```diff
@@ -12,8 +12,10 @@
 
 %if 0%{?rhel}
 %bcond_with zchunk
+%bcond_with zstd
```
dralley, line 15: Unlike zchunk, support for zstd compressed metadata is (I believe) already enabled in libsolv on RHEL.  There's just no convenient way to generate the repos yet.  So it might make sense to likewise default to ON for RHEL >7.  Or at least 9+.

dralley:
I've filed https://github.com/fedora-modularity/libmodulemd/issues/606, as far as I can tell createrepo_c uses its own wrappers for opening up modulemd metadata files so it shouldn't be a problem here, but we wouldn't want to produce zstd compressed module metadata and have libmodulemd not be able to read it.

dralley:
Basic testing with seems to work fine.  Results are good.

Before:

Screenshot from 2023-01-31 23-12-06

After:

Screenshot from 2023-01-31 23-12-37

I'm not sure what the optimal default is but it wouldn't be difficult to tweak later.

j-mracek:
I think it is a great time to change createrepo defaults and not only for compression. But we have to be careful that metadata are compatible with DNF from RHEL8.0

For testing it would be also good to test a large RHEL8 metadata. If I good remember they have a different order of element then Fedora metadata. As most important value I suggest compressed size, decompression time, and RAM requirement for decompression. I am mentioning RAM, because if I good remember with some setting it could create the large dictionary that requires more RAM - minimal system requirements are not small at all, but people have 0.5 or 1 GB RAM on their systems and then they complain to DNF.

I think that the time required to compress files is important but it has less impact then other values.

dralley:
>If I good remember they have a different order of element then Fedora metadata.

You mean no order :)  RHELDST-15791

src/cmd_parser.c
kontura: That is a good idea, I have added that.

createrepo_c.spec
kontura: Ok, I have enabled it also for rhel 8. I believe it was enabled by: https://bugzilla.redhat.com/show_bug.cgi?id=1914876 so I have added a note about that.

kontura:
Ok, I have set it to level 9 for now, it seems like a good compromise between compression efficiency and speed.

I did some quick tests as well (on rhel metadata with 22 000 pkgs) and it seems that zstd (with level 9) is strictly better than gz in all mentioned aspects except maybe for memory usage during decompression but only by a tiny amount (~2%).

dralley (approved):

dralley:
It still looks fine, but what got updated in the last push?

dralley (approved):

kontura:
> It still looks fine, but what got updated in the last push?

Sorry, I should explain, there was a segfault when accessing one of the errors. 
I found it when running https://github.com/rpm-software-management/ci-dnf-stack/pull/1217 test with `mergerepo_c`.

src/cmd_parser.c
```diff
@@ -305,6 +305,8 @@ check_and_set_compression_type(const char *type_str,
         *type = CR_CW_BZ2_COMPRESSION;
     } else if (!strcmp(compress_str->str, "xz")) {
         *type = CR_CW_XZ_COMPRESSION;
+    } else if (!strcmp(compress_str->str, "zstd") || !strcmp(compress_str->str, "zst")) {
```
ppisar, line 308: Why do you recognize non-standard "zstd" file name extension <https://datatracker.ietf.org/doc/html/rfc8878#section-7.1>? If there is no previous use of "zstd" extension in YUM repositories, I would not recognize it. It will only create unneeded pressure to other implementations (e.g. in DNF) to support both extensions "because createrepo_c" does that".

dralley: +1

kontura: That's a good suggestion, I have updated it, thanks.

src/compression_wrapper.c
```diff
@@ -255,6 +277,8 @@ cr_compression_type(const char *name)
         type = CR_CW_XZ_COMPRESSION;
     if (!g_strcmp0(name_lower, "zck"))
         type = CR_CW_ZCK_COMPRESSION;
+    if (!g_strcmp0(name_lower, "zst") || !g_strcmp0(name_lower, "zstd"))
```
ppisar, line 280: What's this "zst" and "zstd" choice? The compression is named "Zstandard" and "zstd". I think you are still mistaken it with "zst" file name extension. Can you remove "zst" option from here?

kontura: yes, sorry, I went through it again, hopefully "zst" is just an extension now.

dralley:
@kontura Is anything remaining blocking this?

kontura:
@dralley it should be ready. I need to find a reviewer.

m-blaha (approved):

PR #348, opened 2023-02-16
Rename filelists_ext to filelists-ext

kontura:
I noticed that it would be more consistent to name the new option `--filelists-ext` and the new file `filelists-ext.xml`.
Follow up for: https://github.com/rpm-software-management/createrepo_c/pull/316
@aplanas what do you think?

It also probably fixes an error where the ext filelists db overrode the normal one.


aplanas:
Yes, it is more consistent. Seems that the fixtures need to be renamed.

Thanks!!

PR #351, opened 2023-02-27
fix errors while parsing utf8 chars in cli options

M0ses:
Without this patch the following error occurs:

"Invalid byte sequence in conversion input"

when you try something like

```
createrepo_c --repo fööbär foobar
```

setlocale is required to ensure that NON-ASCII chars can be parsed SEE https://developer-old.gnome.org/glib/unstable/glib-running.html#setlocale for further information


kontura:
Thank you!

PR #355, opened 2023-04-04
Install distro packages instead of pip ones when available

jan-kolarik:
Currently `Python Release` workflow fails due to conflicting request when a `Pygments` distro package is tried to be uninstalled when pip dependencies are being resolved. In this case we can install the dependencies from distro packages instead.


kontura:
I wonder why it can't use the installed `Pygments` but regardless the change looks good.
Thanks for resolving it.

PR #361, opened 2023-04-26
Add comments to cr_get_header_byte_range()

dralley:
This code is pretty opaque unless you have a good knowledge of the RPM file format, so let's add some comments so it's more clear.

Moved cr_header_byte_range() to parsepkg.c as it's essentially doing package parsing, just manually.  But I can drop that commit if it doesn't make sense.

### DIFF
--- a/src/misc.c
+++ b/src/misc.c
@@ -240,16 +240,30 @@ cr_is_primary(const char *filename)
 }
 */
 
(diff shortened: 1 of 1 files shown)


kontura (commented):
Thanks!

I would prefer not to do the move. Those are public header files and I would rather keep compatibility.

dralley:
@kontura Updated

Issue #360, opened 2023-04-25
Undocumented change in python API


lubomir:
Version 0.21.1 introduced changes in the python API for accessing files in a particular package. The tuple describing a file entry grew from 3 items to 4. I have not found any documentation on what the fourth item is, and there is no `cr.FILE_ENTRY_` constant to access it.

This change breaks backwards compatibility if anyone is using code like:

```python
for type_, path, name in pkg.files:
    ....
```

Here's a small piece of code to replicate the change:
```python
import createrepo_c as cr
md = cr.Metadata()
md.locate_and_load_xml("https://kojipkgs.fedoraproject.org/compose/rawhide/latest-Fedora-Rawhide/compose/Everything/x86_64/os/")
print(len(md.get(md.keys()[0]).files[0]))
```

```
$ rpm -qa | grep createrepo_c
createrepo_c-libs-0.20.1-1.fc36.x86_64
createrepo_c-0.20.1-1.fc36.x86_64
python3-createrepo_c-0.20.1-1.fc36.x86_64
$ python replicator.py 
3
$ sudo dnf update createrepo_c 
...
$ rpm -qa | grep createrepo_c
createrepo_c-libs-0.21.1-1.fc36.x86_64
createrepo_c-0.21.1-1.fc36.x86_64
python3-createrepo_c-0.21.1-1.fc36.x86_64
$ python replicator.py 
4
```

j-mracek:
Thank you for the report. It sounds critical.

kontura:
I made a PR https://github.com/rpm-software-management/createrepo_c/pull/362 to fix this.
Can you please check if everything works well with it for your use cases?

dralley:
@kontura This can be closed now yes?

kontura:
> Should a new issue be filed for my comment on the PR?
> 
> > However to avoid this in the future maybe it makes sense to return a class/dataclass rather than a tuple, so that data can be added without breaking API. Maybe for 1.0? (not backportable obviously)

There is more tuples like this: https://github.com/rpm-software-management/createrepo_c/blob/master/src/python/createrepo_c/__init__.py#L100-L118 plus undocumented `DistroTag`.

I am not sure. 
The change would be quite disruptive but maybe the filelists will be extended again in the future. 

Perhaps we could convert it together with the addition of new elements (if that ever happens).
I think this would be better tracked in code rather than a new issue.

PR #362, opened 2023-04-26
Fix change in python API of package files

kontura:
The change was introduced in https://github.com/rpm-software-management/createrepo_c/pull/316

For: https://github.com/rpm-software-management/createrepo_c/issues/360

### DIFF
--- a/src/python/createrepo_c/__init__.py
+++ b/src/python/createrepo_c/__init__.py
@@ -107,6 +107,7 @@ PCOR_ENTRY_PRE     = 5 #: PCOR entry tuple index - pre
 FILE_ENTRY_TYPE = 0 #: File entry tuple index - file type
 FILE_ENTRY_PATH = 1 #: File entry tuple index - path
 FILE_ENTRY_NAME = 2 #: File entry tuple index - file name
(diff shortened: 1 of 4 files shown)


dralley (approved):
Seems OK to me.  If the digest is only present when you deliberately act to parse a different file, then there shouldn't be a breaking change for existing code.

However to avoid this in the future maybe it makes sense to return a class/dataclass rather than a tuple, so that data can be added without breaking API.  Maybe for 1.0?  (not backportable obviously)

kontura:
> @kontura FYI, we just got minorly bit by this - the API does change for existing code in the event that you parse a file directly - in that case it can return a 4-tuple whereas in other circumstances the API still returns a 3-tuple.
> 
> What's the best way to resolve this, at this point? Or do we just leave it be until the next major release, and work around it?
> 
> [pulp/pulp_rpm#4328](https://github.com/pulp/pulp_rpm/issues/4328)
> 
> IMO, there should at least be a keyword argument that can disable populating the file digests in the returned value. If we cannot fix the API "defaults" otherwise to be consistent, we could at least still do that.

I see, I have also missed that API change. :disappointed: 

Here is a PR that adds a `file_digests` option to `package_from_rpm` to restore the old tuple with 3 members: https://github.com/rpm-software-management/createrepo_c/pull/470.

PR #371, opened 2023-06-07
Switch the default compression type to zstd

kontura:
It also changes mergerepo_c so that option `--compress-type` sets compression for all repository files (except sqlite dbs).

Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1311

Compression for sqlite dbs remains the same.

This is part of: https://fedoraproject.org/wiki/Changes/createrepo_c_1.0.0


jan-kolarik (approved):
LGTM

PR #379, opened 2023-08-10
When compiled without zstd switch default compression back to gz

kontura:
Without this change when compiled without zstd normal run ends in an error because of missing zstd.

Other approach could be to not allow building without zstd.

### DIFF
--- a/src/createrepo_c.c
+++ b/src/createrepo_c.c
@@ -826,9 +826,14 @@ main(int argc, char **argv)
     const char *xml_compression_suffix = NULL;
     const char *sqlite_compression_suffix = NULL;
     const char *compression_suffix = NULL;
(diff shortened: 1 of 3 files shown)


dralley (approved):

Conan-Kudo:
> Other approach could be to not allow building without zstd.

I would rather us make zstd non-optional.

j-mracek:
I also think that we could remove the condition for zstd. It was there for old RHEL and we have no plans for rebase there. Createrepo_c is not in the minimal deployment therefore additional dependency is not an issue.

asarubbo:
> > Other approach could be to not allow building without zstd.
> 
> I would rather us make zstd non-optional.

See #383

kontura:
I am closing this in favor of: https://github.com/rpm-software-management/createrepo_c/pull/392

Issue #384, opened 2023-09-11
heap buffer overflow and stack buffer overflow in test suite


asarubbo:
Our [Gentoo tinderbox](https://blogs.gentoo.org/ago/2020/07/04/gentoo-tinderbox/) reported at [bug 913966](https://bugs.gentoo.org/913966) a test failure.

Looking at the log attached to the downstream report, we can see:

```
==25==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6060000651b8 at pc 0x56273fab2a32 bp 0x7fffedf00ee0 sp 0x7fffedf00688
READ of size 65 at 0x6060000651b8 thread T0
    #0 0x56273fab2a31 in __interceptor_strlen.part.0 (/var/tmp/portage/app-arch/createrepo_c-1.0.0/work/createrepo_c-1.0.0_build/tests/test_compression_wrapper+0x63a31)
    #1 0x56273fb7cf0a in test_cr_get_zchunk_with_index /var/tmp/portage/app-arch/createrepo_c-1.0.0/work/createrepo_c-1.0.0/tests/test_compression_wrapper.c:903
    #2 0x7fd72023e8ad  (/usr/lib64/libglib-2.0.so.0+0x838ad)
    #3 0x7fd72023e6a2  (/usr/lib64/libglib-2.0.so.0+0x836a2)
    #4 0x7fd72023edc1 in g_test_run_suite (/usr/lib64/libglib-2.0.so.0+0x83dc1)
    #5 0x7fd72023ee47 in g_test_run (/usr/lib64/libglib-2.0.so.0+0x83e47)
    #6 0x56273fa71687 in main /var/tmp/portage/app-arch/createrepo_c-1.0.0/work/createrepo_c-1.0.0/tests/test_compression_wrapper.c:953
    #7 0x7fd71fe23c89  (/lib64/libc.so.6+0x23c89)
    #8 0x7fd71fe23d44 in __libc_start_main (/lib64/libc.so.6+0x23d44)
    #9 0x56273fa71750 in _start (/var/tmp/portage/app-arch/createrepo_c-1.0.0/work/createrepo_c-1.0.0_build/tests/test_compression_wrapper+0x22750)

0x6060000651b8 is located 0 bytes after 56-byte region [0x606000065180,0x6060000651b8)
allocated by thread T0 here:
[...]

PR #385, opened 2023-09-21
Add `sanitizers` build option, fix bunch of detected problems

kontura:
This should fix most of the detected problems. There still are several left that I wasn't able to figure out and I am not sure if they are problems in createrepo_c or if they are not false positives:
- Leaks in python unittest (this might be a python loading module problem - we see something like this even in dnf)
- Leaks in `g_option_context_parse`
- Inconsistent leaks from `xmlBufferCreate` from `cr_xml_dump_primary`
- Leak of `sqliterepocmdoptions_new`

Closes: https://github.com/rpm-software-management/createrepo_c/issues/384

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,6 +111,12 @@ if (WITH_ZSTD)
     SET (CMAKE_C_FLAGS_DEBUG    "${CMAKE_C_FLAGS_DEBUG} -DWITH_ZSTD")
 endif()
 
+if (WITH_SANITIZERS)
+    message(WARNING "Building with sanitizers enabled!")
+    add_compile_options(-fsanitize=address -fsanitize=undefined -fsanitize=leak)
+    link_libraries(asan ubsan)
+endif()
+
 # Threaded XZ Compression
 # Note: This option is disabled by default, because Createrepo_c
 # parallelize a lot of tasks (including compression) by default, this
--- a/createrepo_c.spec
+++ b/createrepo_c.spec
@@ -31,6 +31,8 @@
 %bcond_with legacy_hashes
 %endif
 
+%bcond_with sanitizers
+
 Summary:        Creates a common metadata repository
 Name:           createrepo_c
 Version:        1.0.0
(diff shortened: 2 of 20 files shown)


dralley:
I'm still catching errors when publishing metadata.

`.....Error: Task /pulp/api/v3/tasks/018af72f-89dd-7562-9b44-2581415dc286/ failed: 'Killed by signal 11.'`

Unfortunately it's running inside a container, and dumps the coredump on my host system, where I don't have the debug symbols to inspect it properly.  I'm sure it's possible but I need to figure out how exactly.

dralley:
@kontura I pinned it down to the code which sets `files` on the package object, on the Python side.

Here's a minimal reproducer

```
In [1]: import createrepo_c as cr

In [2]: pkg = cr.Package()

In [3]: pkg.files = [(None, '/etc/accel-config/', 'accel-config.conf.sample')]
Segmentation fault (core dumped)

```

It appears that all of the [tests](https://github.com/rpm-software-management/createrepo_c/blob/master/tests/python/tests/test_package.py) which exercise that setter use an empty string in the first position of the tuple whereas we pass `None`.  But, `None` is also the value that createrepo_c will return from parsed metadata, so if you take parsed metadata and then set `pkg.files = pkg.files` on the resulting objects then it will also crash.

Also I thought [this](https://github.com/rpm-software-management/createrepo_c/commit/b254b449079ae72d94a266221e2a5e357dd0daf3#diff-4e6fd12cb53827103ebca4de64943c383b6206bab761127864ded871dc409eddR110) looked a bit suspicious, since it was unconditional, I thought maybe it could be a `free(NULL)`.   It isn't the cause of the crash so I haven't looked further.

dralley:
If you can put out a release with this fixed as soon as possible once this is merged, that would be appreciated.  We just got our first bug report in Satellite that a repo which has switched to zstd metadata is un-syncable, so we'll need to pull the new release and have it backported to many Satellite releases most likely.

kontura:
@dralley Thanks for the details.
I made a PR: https://github.com/rpm-software-management/createrepo_c/pull/388

Once both it and this PR are merged I will make a new release.
I consider both quite important.

dralley (approved):

dcantrell (approved):

PR #387, opened 2023-09-26
Move cr_compress_groupfile outside WITH_LIBMODULEMD

kraj:
This function is used in code which is not conditional under WITH_LIBMODULEMD therefore the declaration should also match its definition scope

Fixes build issues flagged by clang

src/createrepo_c.c:850:16: error: incompatible integer to pointer conversion initializing 'gchar *' (aka 'char *') with an
 expression of type 'int' [-Wint-conversion]
|   850 |         gchar *compressed_path = cr_compress_groupfile(cmd_options->groupfile_fullpath, tmp_out_repo, compression);
|       |                ^                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


kontura (approved):
Thank you!

PR #390, opened 2023-10-13
Fix building on EL9

dralley:
The existing code is non-idiomatic, and as it so happens EL9 currently defines this macro unexpectedly despite not having the proper API, so we replace this with the idiomatic approach to make it actually work.

https://issues.redhat.com/browse/RHEL-12941
https://github.com/rpm-software-management/createrepo_c/pull/342/files#r1357391661

### DIFF
--- a/src/createrepo_c.c
+++ b/src/createrepo_c.c
@@ -63,7 +63,7 @@
  * Starting with glib 2.70.0, g_pattern_spec_match() replaces
  * g_pattern_match().
  */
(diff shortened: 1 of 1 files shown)


mcatanzaro:
Looks good!

dralley:
@kontura

kontura (approved):

oyvindkolbu:
Tested this locally on EL9, Rocky 9 in my case and worked fine. Would be great with a new release with it. Thank you!

Issue #353, opened 2023-03-24
Auto-detect compression appears to be broken for some files


dralley:
If a file is downloaded to a temporary file, it may not necessarily retain the original extension.  I would still expect the magic number to be recognized but I'm seeing some *really* weird behavior with this.

Below I've downloaded the primary.xml.gz from two repos:

- https://dl.fedoraproject.org/pub/fedora/linux/updates/37/Everything/x86_64/repodata/
- https://download.copr.fedorainfracloud.org/results/audron/dexed/fedora-38-aarch64/repodata/

I've also renamed the file to remove the file extension on the 2nd attempt.  Strangely you can see that for one file it still works, and for the other file it doesn't.

```
In [1]: import createrepo_c as cr

In [2]: cr.detect_compression("fbe45ef42039fec552386d5c18de2175d7c525994ae20aed7fedd4a990891da7-prim
   ...: ary.xml.gz")
Out[2]: 3

In [4]: cr.detect_compression("fbe45ef42039fec552386d5c18de2175d7c525994ae20aed7fedd4a990891da7")
Out[4]: 3

In [5]: cr.detect_compression("af12533499469117e3824bc938bfc3208539deb44bbf56ec2612fc383b28a1f4-prim
   ...: ary.xml.gz")
Out[5]: 3

In [6]: cr.detect_compression("af12533499469117e3824bc938bfc3208539deb44bbf56ec2612fc383b28a1f4")
Out[6]: 1
```

Using the `file` command directly shows the same:

```
[dalley@thinkpad Downloads]$ file -i fbe45ef42039fec552386d5c18de2175d7c525994ae20aed7fedd4a990891da7 
fbe45ef42039fec552386d5c18de2175d7c525994ae20aed7fedd4a990891da7: application/gzip; charset=binary
[dalley@thinkpad Downloads]$ file -i af12533499469117e3824bc938bfc3208539deb44bbf56ec2612fc383b28a1f4 
af12533499469117e3824bc938bfc3208539deb44bbf56ec2612fc383b28a1f4: application/octet-stream; charset=binary
```

But the `gzip` magic number `1f8b` exists in both files?  This is extremely strange.  Why does file have problems with one but not the other?

```
[dalley@thinkpad Downloads]$ xxd fbe45ef42039fec552386d5c18de2175d7c525994ae20aed7fedd4a990891da7 | head
00000000: 1f8b 0800 0000 0000 0003 ec7d 5b97 db36  ...........}[..6
[...]

PR #391, opened 2023-10-15
Remove dependency on libmagic, hardcode magic number detection, build binary packages for PyPI

dralley:
#145 is blocked by the fact that createrepo_c depends on libmagic, which references a system-provided database file.  In the event that createrepo_c is built statically the version of libmagic baked into the executable may not be able to read this database file, presuming it is of a newer or older format.

Given we only need to understand a small number of compression types, we can dispense with libmagic and detect them based on the magic number ourselves.


dralley:
The current test failure is

```
        # Bad suffix - no compression
        path = os.path.join(COMPRESSED_FILES_PATH, "01_plain.foo0")
        comtype = cr.detect_compression(path)
>       self.assertEqual(comtype, cr.NO_COMPRESSION)
E       AssertionError: 3 != 2
```

Is it a hard requirement that we return `cr.NO_COMPRESSION` even when there is no suffix and no magic number to aid detection?  It seems to me that there's no great way to distinguish `NO_COMPRESSION` vs. `UNKNOWN_COMPRESSION` without trying to duplicate all of the heuristics libmagic uses (which even then, would [not be perfect](https://github.com/rpm-software-management/createrepo_c/issues/353#issuecomment-1485094556)).

dralley:
Well, I added a heuristic anyway.  I don't love it but I doubt it's any worse than whatever libmagic does.  It's impossible to be perfect in this case.

Personally I would prefer to fall back to `NO_COMPRESSION` unless there is an extension we recognize, wrapped in another extension we don't.

tests/test_compression_wrapper.c
```diff
@@ -27,7 +27,7 @@
 #include "createrepo/misc.h"
 #include "createrepo/compression_wrapper.h"
 
-#define TMP_FILE_PATTERN                        "test_XXXXXX.txt"
```
dralley, line 30: This broke when I added this line https://github.com/rpm-software-management/createrepo_c/pull/391/files#diff-c87d8ed02804b9ad227639fcb6469f0bba1256da4cf73f1937a6d023bfd374b9R179

dralley:
Not really sure why the modulemd merging test is failing

src/compression_wrapper.c
```diff
@@ -171,91 +175,59 @@ cr_detect_compression(const char *filename, GError **err)
...
-        else if (g_str_has_prefix(mime_type, "application/x-xz"))
-        {
-            type = CR_CW_XZ_COMPRESSION;
+    fclose(file);
+
+    // assume that most compressed files will have a high bit set somewhere in the first 5 bytes
+    // and most uncompressed files (generally text) will not
+    bool is_text_heuristic = true;
+    for (int i=0; i<bytesRead; i++) {
+        if (magic[i] > 127) {
```
kontura, line 207: I believe this comparison is always `false`, since `magic` is a `char` array and `char` has a range of -128 to 127 .

src/compression_wrapper.c
```diff
@@ -171,91 +175,59 @@ cr_detect_compression(const char *filename, GError **err)
...
-
-        else if (g_str_has_prefix(mime_type, "application/x-xz"))
-        {
-            type = CR_CW_XZ_COMPRESSION;
+    fclose(file);
+
+    // assume that most compressed files will have a high bit set somewhere in the first 5 bytes
+    // and most uncompressed files (generally text) will not
+    bool is_text_heuristic = true;
+    for (int i=0; i<bytesRead; i++) {
```
kontura, line 206: Since `bytesRead` is a `size_t` can you make also `i` a `size_t`?
So that we don't compare `unsigned` with `signed`.

src/compression_wrapper.c
```diff
@@ -132,11 +131,16 @@ typedef struct {
 } ZstdFile;
 #endif
 
+void bufferToHexString(const unsigned char* buffer, int size, char* hexString, int hexStringSize) {
```
kontura, line 134: Since we are passing the `magic` into this can we match the `buffer` and `magic` types? 
For example to just `const char* buffer`.

kontura:
I am also not a fan of the heuristics.

Yea, when we cannot recognize the extension or magic bytes I would fall back to `NO_COMPRESSION` and leave the `UNKNOWN_COMPRESSION` basically for the errors.

It is a change and the tests will need to be amended but I would be ok with that.

dralley:
@kontura Ready for another look

dralley:
Added the machinery for building wheels on release.

dralley:
I think the COPR builders are down.

kontura (approved):
Thank you!

kontura:
@dralley I made a new release and the PyPI publishing failed: https://github.com/rpm-software-management/createrepo_c/actions/runs/6689206207/job/18172463650

Is it because we are missing a `long_description` in `setup.py`?

dralley:
@kontura I changed that and ran the commands locally and was able to upload the packages, so I suppose that works yeah.

PR #392, opened 2023-10-19
Don't allow building without zstd

kontura:
Since zstd is the default compression building without leads to broken default runs.

Alternative to: https://github.com/rpm-software-management/createrepo_c/pull/379


j-mracek:
LGTM

PR #395, opened 2023-11-08
Add a --no-pretty option to reduce repo metadata size

dirkmueller:
The indentation might be pretty, but it also adds a few percent of compressed size to the metadata. for production use a --no-pretty can now be set that reduces the amount of blanks in the repodata xml.

### DIFF
--- a/doc/createrepo_c.8
+++ b/doc/createrepo_c.8
@@ -85,6 +85,9 @@ Choose the checksum type used in repomd.xml and for packages in the metadata. Th
 .SS \-p \-\-pretty
 .sp
 Make sure all xml generated is formatted (default)
+.SS \-p \-\-no\-pretty
+.sp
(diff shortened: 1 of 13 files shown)


Conan-Kudo:
Unless you're not compressing the metadata files, this doesn't really make sense. Also, people do download and look at those files, so I'd rather not make it completely unreadable for just a percentage point or two.

dirkmueller:
@Conan-Kudo I am comparing compressed sizes. The gain would be much bigger in raw. Also, the default does not change. It just adds an option to change it.

The files are not completely unreadable. Basically one package is one line in the output instead of multiple lines. Plus it's just a `xmliint --format`away to get the previous output.

Conan-Kudo:
Keep in mind that this doesn't affect anything that is added to the repodata through `modifyrepo`, but sure, I guess...

Conan-Kudo (approved):

Issue #404, opened 2023-11-21
sqlite3_enable_shared_cache


tpgxyz:
Hi, looks like there is no check if SQLite is compiled with -DSQLITE_OMIT_SHARED_CACHE
According to SQLite upstream docs this option is recommended to use https://www.sqlite.org/c3ref/enable_shared_cache.html

[sqlite3_enable_shared_cache](https://github.com/rpm-software-management/createrepo_c/blob/c10ab17792d64f12a448dfb0dea96afa39e7d24d/src/sqlite.c#L1542)

I’d surges to make this sqlite3_enable_shared_cache  call conditional if SQLite is compiled with -DSQLITE_OMIT_SHARED_CACHE

ppisar:
Thanks for the report. I'd rarher replace it together with sqlite3_open() with sqlite3_open_v2(,, SQLITE_OPEN_READWRITE|SQLITE_OPEN_SHAREDCACHE, NULL) call. Not only the flags works is ignored by sqlite builds compiled with -DSQLITE_OMIT_SHARED_CACHE, it also prevents from contaminating the shared cache use for SQLite connections not opend by createrepo_c library.

SQLITE_OPEN_SHAREDCACHE flag was added in sqlite-3.6.18 4 years ago.

PR #405, opened 2023-11-23
SQLite: Enable shared cache on connection level

ppisar:
SQLite shared cache is discouraged and SQLite can be compiled without a support for it. If the support is disabled,
sqlite3_enable_shared_cache() function is not provided by the library and building createrepo_c fails.

This patch moves from sqlite3_enable_shared_cache() to a newer sqlite3_open_v2(,, SQLITE_OPEN_SHAREDCACHE,) API available since SQLite 3.6.18 (2009-09-11).

This has four effects:

(1) Building passes even if shared cache is not supported by SQLite.

(2) If the support is not available at run-time, SQLite keeps working.

(3) The shared cache is only enabled on connections createrep_c library opens. It does not affect SQLite connections made by different libraries or an application.

(4) sqlite >= 3.6.18 is required for building.

https://github.com/rpm-software-management/createrepo_c/issues/404 https://www.sqlite.org/c3ref/enable_shared_cache.html


kontura (approved):
Thanks!

PR #406, opened 2023-11-24
Use Epoch-friendly sub-package interdependencies

ppisar:
This helps with minimizing downstream-patches for custom createrepo_c builds with bumped Epoch (useful for the Copr team, we need to maintain a custom createrep_c build with legacy hashes enabled, and bumped epoch protects us from random upgrades to the system-default version).


ppisar:
This comes from Fedora <https://src.fedoraproject.org/rpms/createrepo_c/pull-request/14>.

m-blaha (approved):
Thanks!

PR #408, opened 2023-11-24
Fix const-correctness with libxml2-2.12.0

ppisar:
libxml2-2.12.0 started to return errors as a pointer to a constant structure and GCC 13.2.1 started to warn like this:

/home/test/createrepo_c/src/xml_parser.c: In function ‘cr_xml_parser_generic’: /home/test/createrepo_c/src/xml_parser.c:204:35: warning: initialization discards ‘const’ qualifier fro m pointer target type [-Wdiscarded-qualifiers]
  204 |             xmlErrorPtr xml_err = xmlCtxtGetLastError(parser);
      |                                   ^~~~~~~~~~~~~~~~~~~

This patch fixes it.


kontura (approved):
Thank you.

PR #411, opened 2023-12-02
Use a gzip compression by default when --compatibility is specified

dralley:
Unless the user explicitly asks for a different compression algo.


kontura:
Thanks!

PR #412, opened 2023-12-05
Update README

dralley:


README.md
```diff
@@ -331,7 +331,8 @@ Modifyrepo - at least not yet.
 | type          | Type of the metadata | Any string | Based on filename |
 | remove        | Remove specified file/type from repodata | ``true`` or ``false`` | ``false`` |
 | compress      | Compress the new metadata before adding it to repo | ``true`` or ``false`` | ``true`` |
-| compress-type | Compression format to use | ``gz``, ``bz2``, ``xz`` | ``gz`` |
+| compress-type | Compression format to use for auxillary metadata | ``gz``, ``zstd``, ``bz2``, ``xz`` | ``zstd`` |
+| general-compress-type | Compression format to use for primary metadata | ``gz``, ``zstd``, ``bz2``, ``xz`` | ``zstd`` |
```
kontura, line 335: These options are for the [modifyrepo_c batchfile](https://github.com/rpm-software-management/createrepo_c/blob/master/src/modifyrepo_shared.c#L499-L519) I don't think it will use the `general-compress-type`.
The `compress-type` is also not specific to `auxillary metadata` but to the modified metadata.

PR #416, opened 2024-01-01
Fix compression help text, zck handling

direc85:
I started using createrepo_c recently. I noticed a typo in the help text, so I thought I'd fix it. It looked like zck handling was missing altogether, so I added that in, too.

- Fix help text mixing up file extensions and compression methods
- Add zck method handling
- Put zck help text and parsing in `#ifdef WITH_ZCHUNK`

### DIFF
--- a/src/cmd_parser.c
+++ b/src/cmd_parser.c
@@ -182,16 +182,20 @@ static GOptionEntry cmd_entries[] =
       "Number of workers to spawn to read rpms.", NULL },
     { "xz", 0, 0, G_OPTION_ARG_NONE, &(_cmd_options.xz_compression),
       "Use xz for repodata compression.", NULL },
-    { "compress-type", 0, 0, G_OPTION_ARG_STRING, &(_cmd_options.compress_type),
-      "Which compression type to use for additional metadata files (comps, updateinfo, etc). Supported compressions are: bz2, gz, zck, zstd, xz.", "COMPRESSION_TYPE" },
-    { "general-compress-type", 0, 0, G_OPTION_ARG_STRING, &(_cmd_options.general_compress_type),
-      "Which compression type to use (even for primary, filelists and other xml). Supported compressions are: bz2, gz, zck, zstd, xz.",
-      "COMPRESSION_TYPE" },
 #ifdef WITH_ZCHUNK
+    { "compress-type", 0, 0, G_OPTION_ARG_STRING, &(_cmd_options.compress_type),
+      "Which compression type to use for additional metadata files (comps, updateinfo, etc). Supported values are: bz2, gz, zck, zstd, xz.", "COMPRESSION_TYPE" },
+    { "general-compress-type", 0, 0, G_OPTION_ARG_STRING, &(_cmd_options.general_compress_type),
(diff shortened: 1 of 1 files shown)


src/cmd_parser.c
```diff
@@ -318,7 +323,11 @@ check_and_set_compression_type(const char *type_str,
         *type = CR_CW_BZ2_COMPRESSION;
     } else if (!strcmp(compress_str->str, "xz")) {
         *type = CR_CW_XZ_COMPRESSION;
-    } else if (!strcmp(compress_str->str, "zstd")) {
```
kontura, line 321: This would be a breaking change so we cannot do that. 
I am not sure if the intention was to specify the compression by its suffix but I would keep it as `zstd`.

kontura (changes requested):
I think that people usually generate the zchunk metadata in addition to some other compression using the `--zck` option but it does make sense to me to allow specifying only the zck compression.

Thank you for the fixes.

ppisar:
Please remove the rename of "zstd" to "zst" from the pull request. Otherwise, it looks good.

direc85:
While I aim for correctness, I do agree that a breaking change should not be introduced over this. As long as the `--help` text agrees with the actual code, I'm happy :)

I'll remove the breaking change part (zst) and amend the pull request next.

Thanks for the feedback!

src/cmd_parser.c
```diff
@@ -182,16 +182,21 @@ static GOptionEntry cmd_entries[] =
...
     { "zck", 0, 0, G_OPTION_ARG_NONE, &(_cmd_options.zck_compression),
       "Generate zchunk files as well as the standard repodata.", NULL },
     { "zck-dict-dir", 0, 0, G_OPTION_ARG_FILENAME, &(_cmd_options.zck_dict_dir),
       "Directory containing compression dictionaries for use by zchunk", "ZCK_DICT_DIR" },
+#else
+    { "compress-type", 0, 0, G_OPTION_ARG_STRING, &(_cmd_options.compress_type),
+      "Which compression type to use for additional metadata files (comps, updateinfo, etc). Supported values are: bz2, gz, zstd, xz.", "COMPRESSION_TYPE" },
+    { "general-compress-type", 0, 0, G_OPTION_ARG_STRING, &(_cmd_options.general_compress_type),
+      "Which compression type to use (even for primary, filelists and other xml). Supported values are: bz2, gz, zstd, xz.", "COMPRESSION_TYPE" },
+
```
ppisar, line 199: Just a nitpick: Please remove this new empty line.

ppisar:
"createrepo_c --general-compress-type=zck --zck" fails when compressing the files:

~~~~
Directory walk started
Directory walk done - 1 packages
Temporary output repo path: ./.repodata/
Critical: Cannot open file ./.repodata//primary.xml.zck: File already exists
~~~~

Should we reject --zck option with --general-compress-type=zck option sooner, when parsing options? E.g. in check_arguments() near already existing "Cannot use --zck-dict-dir without setting --zck" message? Could you add the rejection to your commit?

ppisar (approved):
Now it's good.

PR #418, opened 2024-02-06
Extend the --compatibility flag to cover other defaults

dralley:
* --compatibility will enable sqlite metadata by default
* --compatibility will leave group metadata uncompressed rather than defaulting to the same compression as everything else

I believe this covers the bases in terms of ensuring createrepo_c 1.0 can still generate repos consumable by ancient RHEL / SUSE without too much effort.

I don't add  `group_gz` (separate gzip-compressed group metadata) back - this is aiming for minimal additional complexity and `group_gz` simply isn't worth it.  It's usually a relatively tiny metadata file anyway.


dralley:
We got a complaint from a user that was using `createrepo_c` on their Satellite box.  They're not really supposed to do that, but since Koji and others have ran into the same issue, might as well just implement the simple workaround.

kontura (approved):
Thank you!

dralley:
@kontura Could you do a quick release w/ these changes?

PR #419, opened 2024-02-06
mergerepo_c and modifyrepo_c now make "pretty" indented xml

### LINKED ISSUE #413: `modifyrepo_c` and `mergerepo_c` generate `--no-pretty` metadata by default
Since https://github.com/rpm-software-management/createrepo_c/pull/395 both `modifyrepo_c` and `mergerepo_c` generate metadata without full indentation.

This is because the default to full indentation is set in `cr_xml_dump_init()` but neither of the tools calls that.

dralley:
closes #413

### DIFF
--- a/src/mergerepo_c.c
+++ b/src/mergerepo_c.c
@@ -875,6 +875,10 @@ dump_merged_metadata(GHashTable *merged_hashtable,
 {
     GError *tmp_err = NULL;
 
+    // Set up XML dump parameters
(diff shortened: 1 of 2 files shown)


dralley:
Removed "by default" from the title because these tools don't have `--pretty` / `--no-pretty` flags.

kontura (approved):
Thanks!

Issue #403, opened 2023-11-20
Newer createrepo_c doesn't generate comps readable EL7


kwizart:
I'm using an (older) fedora with koji to build from el7 to fedora rawhide (as the rpmfusion.org project)

Using a backported createrepo_c 1.0.2-2 , the command generate a repomd.xml with a compressed comps.xml, but this file cannot be read by yum that seems to expect an uncompressed comps.xml
```
/usr/bin/createrepo_c --general-compress-type=gz -vd -o /var/tmp/koji/tasks/8390/618390/repo -i /mnt/koji/repos/el7-free-build/80086/x86_64/pkglist -g /mnt/koji/repos/el7-free-build/80086/groups/comps.xml --update --skip-stat /mnt/koji/repos/el7-free-build/80086/x86_64
```

Even using the compatibility mode (using createrepo or createrepo_c --compatibility) doesn't change this behavior.

When the koji builder populates the buildroot (for the initial buildfromSCM buildroot), I'm experiencing the following error:
```
DEBUG util.py:443:  There is no installed groups file.
DEBUG util.py:443:  Maybe run: yum groups mark convert (see man yum)
DEBUG util.py:443:  Failed to add groups file for repository: build - comps file is empty/damaged
DEBUG util.py:443:  Warning: group srpm-build does not exist.
DEBUG util.py:443:  Maybe run: yum groups mark install (see man yum)
DEBUG util.py:443:  Error: No packages in any requested group available to install or update

```

My current workaround is to manually uncompress the comps file and modify the repodata.xml as appropriate.
After this modification, the build can occurs normally.

kwizart:
Side note, I had to use --general-compress-type=gz (but I plan to use =xz) because I expect zstd will not work on EL7.
This is patched kojid, but I expect a sane default should be picked as koji is unlikely to have a per build target compression format...

dralley:
>Using a backported createrepo_c 1.0.2-2

But why?  Is there any reason to use such a new version of createrepo_c with a distro that will be out of support in 6 months?

kwizart:
I'm sorry, I think you miss-understood, I'm not using createrepo_c {on} el7, I'm using to generate repodata {for} el7!

sergiomb2:
[libcomps](https://koji.fedoraproject.org/koji/packageinfo?packageID=16769) was updated recently (one month ago) , can be related ?

kwizart:
[...]

PR #420, opened 2024-02-07
Use compatibility for createrepo

kwizart:
Reported as #403

The point of using createrepo is to keep the same behavior with the createrepo command. Users that want modern behavior should use createrepo_c command instead.

### DIFF
--- a/src/createrepo_c.c
+++ b/src/createrepo_c.c
@@ -634,6 +634,11 @@ main(int argc, char **argv)
         }
     }
 
+    // Using createrepo assumes compatibility mode
+    if (!g_strcmp0(g_path_get_basename(argv[0]),"createrepo")) {
+        cmd_options->compatibility=TRUE;
+    }
+
     // Dirs
     gchar *in_dir       = NULL;  // path/to/repo/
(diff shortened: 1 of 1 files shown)


kwizart:
Tested on my side with this koji scratch build for fedora rawhide
https://koji.fedoraproject.org/koji/taskinfo?taskID=113132488

Conan-Kudo (changes requested):
I would rather not have this. Tools should adjust accordingly instead.

kwizart:
Do you have a particular reason why you want to change every dependent software instead ?
And can you also include what's the point to keep different commands createrepo and createrepo_c if they have the same behavior ?

Conan-Kudo:
Ideally, we wouldn't produce the `_c` binaries anymore. There's no point to them anymore since the old py-createrepo is gone.

kwizart:
Well, I'm just exactly raising a point for a createrepo use over createrepo_c.

dralley:
IMO, this doesn't feel particularly worthwhile at this point

kontura:
Right, there is the `--compatibility` flag.

PR #422, opened 2024-02-21
Build aarch64 python packages during releases

dralley:


.github/workflows/release-python.yml
```diff
@@ -36,6 +36,15 @@ jobs:
   build_bdist:
     name: Build binary wheels
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - arch: auto64  # native 64-bit
+            skip: "pp* *-musllinux_* cp36* cp37* cp38*"  # no PyPy or musl builds, no older Python versions
+          - arch: aarch64
+            skip: "pp* *-musllinux_* cp36* cp37* cp38* cp310* cp312*"   # qemu builds are incredibly slow, so only do 3.9 and 3.11
```
dralley, line 45: The four builds "only" take 45 minutes or so under emulation, as opposed to like 3 minutes normally, so if you would prefer we can enable builds for all versions.

inknos: I don't see an issue with the time increase since it will run on a release hook

inknos: I would merge it for now and make a note to check the status on the next createrepo_c release to see how it goes. Thanks

PR #432, opened 2024-05-30
spec: Package Python examples

ppisar:
We already use it in F41 package. I'm only not sure since when %py3_shebang_fix is supported.

### DIFF
--- a/createrepo_c.spec
+++ b/createrepo_c.spec
@@ -127,7 +127,7 @@ Python 3 bindings for the createrepo_c library.
 
 %prep
 %autosetup -p1
-
+%py3_shebang_fix examples/python
 mkdir build-py3
 
 %build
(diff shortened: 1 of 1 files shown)


j-mracek:
Macro is available on Fedora 39 - tested by `rpm --eval '%{py3_shebang_fix}'`

ppisar:
I found that macro in RHEL 9 and 8. I was unable to check RHEL 7.

j-mracek:
I think that this option is better then to replace `#!/usr/bin/python`. But may I ask you to add why you want to change the code?

The current code requires to have installed `python-unversioned-command` to make example functional.

ppisar:
I use the macro because all packaged files needs to conform the packaging guidelines. /usr/bin/env as found in the sources does not conform and rpmlint correctly warns on it. 

Do you want me to add this explanation to the commit message?

What do you mean with "python-unversioned-command"? I cannot see it anywhere in the sources, neither in the built <https://kojipkgs.fedoraproject.org//packages/createrepo_c/1.1.2/1.fc41/x86_64/python3-createrepo_c-1.1.2-1.fc41.x86_64.rpm>. 

What do you mean with "The current code requires [...] to make example functional"? Documentation files is exempted from adding RPM dependencies.

j-mracek:
Please may I ask you to add to commit message that the change is required by Fedora guidelines with a reference?

My previous commend about `python-unversioned-command` package was related to `#!/usr/bin/python` but in `createrepo` there is `#!/usr/bin/env python`, therefore not related at all. Apologizing.

ppisar:
Done.

j-mracek (approved):
LGTM

PR #434, opened 2024-07-01
Fix a file descriptor and memory leak in an error path of cr_detect_compression()

ppisar:
Covscan reported:

    createrepo_c-1.1.2/src/compression_wrapper.c:197: error[resourceLeak]: Resource leak: file

It's a real bug and this patch fixes it.

Resolves: https://issues.redhat.com/browse/RHEL-45645


m-blaha (approved):
Thank you!

Issue #436, opened 2024-09-23
Failing of running make_rpm.sh


InHavk:
Hello,
I tried to build rpm for actual release (1.1.4) via utils/make_rpm.sh and got some problems with it:
- utils/make_rpm.sh ignores reference argument
- spec file try to find tar.gz, but make_tarball generate tag.xz
- make_tarball don't append revision in directory name, which is required for spec

OS: Almalinux 9.4

I made fixes (will attach PR after), but i'm not sure that it isn't broke any other scenarios. So, I'm ready to discussion.

InHavk:
remake it a little bit: instead of adding version to directory name, set spec to ignore that

PR #437, opened 2024-09-23
Fix rpm build script

InHavk:
Reported as #436 

Some small fixes of generating rpm packages

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ snippets/
 # Packages and tarball
 createrepo*.rpm
 python-createrepo*.rpm
+python3-createrepo*.rpm
 createrepo_c-*.tar.xz
 
 # Makefile generated files
--- a/createrepo_c.spec
+++ b/createrepo_c.spec
@@ -32,9 +32,15 @@
 
 %bcond_with sanitizers
 
+%if %{defined gitrev}
+%define package_version %{?gitrev}
+%else
+%define package_version 1.2.0
+%endif
+
 Summary:        Creates a common metadata repository
 Name:           createrepo_c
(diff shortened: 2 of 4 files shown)


kontura (changes requested):
Can you drop the `spec` file changes?
We keep this `spec` in sync with fedora downstream createrepo_c `spec` files and this would break our workflow. We are using the Github releases and those are `tar.gz`.

I would rather you update to `make_tarball.sh`.

InHavk:
Sorry for long waiting, I've lost notification.

Problem there not only with format of archive. That part is easy to fix without touching spec.

Source tar named as a `%{name}-%{version}.tar.gz` in spec and `version` is hardcoded also there. It means, that spec always will wait `createrepo_c-1.1.4.tar.gz` with `createrepo_c-1.1.4` directory inside.

So, there is two ways:
1 - We can ignore `refs` as a version and `make_tarball.sh` always will make tar only with version that exact same as in spec. For instance, `make_tarball.sh branchname` will use `branchname` as source of code, but have `createrepo_c-1.1.4.tar.gz` in result.
2 - Add in spec some variable like `GITREV` with condition like `if GITREV exist - use it as version, instead use hardcoded version`. It shouldn't break building rpm in any other situation, if that variable hasn't used.

kontura:
I would find the condition is spec file less confusing so I prefer the second approach.

kontura (approved):
I am sorry I didn't realize there will be a conflict and I bumped the version.
Can you please rebase?

Otherwise it looks good.

InHavk:
Yeah, no problem. Rebased

kontura:
Thank you

PR #438, opened 2024-10-01
Use IMPORTED_TARGET for 3rd-party dependencies

gahr:
The current CMakeLists.txt fails to include the required link directories for 3rd-party packages. As an example, on FreeBSD where packages are installed under /usr/local, the link lines include -lgio-2.0 but not -L/usr/local/lib.

The suggested solution is to use the IMPORTED_TARGET mode of pkg_check_modules. This requires CMake 3.6, so I have bumped the minimum required version.


src/CMakeLists.txt
```diff
@@ -86,18 +86,18 @@ ENDIF ()
...
-TARGET_LINK_LIBRARIES(libcreaterepo_c ${RPM_LIBRARIES})
-TARGET_LINK_LIBRARIES(libcreaterepo_c ${SQLITE3_LIBRARIES})
+TARGET_LINK_LIBRARIES(libcreaterepo_c PkgConfig::RPM)
+TARGET_LINK_LIBRARIES(libcreaterepo_c PkgConfig::SQLITE3)
 TARGET_LINK_LIBRARIES(libcreaterepo_c ${ZLIB_LIBRARY})
-TARGET_LINK_LIBRARIES(libcreaterepo_c ${ZCK_LIBRARIES})
-TARGET_LINK_LIBRARIES(libcreaterepo_c ${DRPM_LIBRARIES})
-TARGET_LINK_LIBRARIES(libcreaterepo_c ${ZSTD_LIBRARIES})
+TARGET_LINK_LIBRARIES(libcreaterepo_c PkgConfig::ZCK)
+TARGET_LINK_LIBRARIES(libcreaterepo_c PkgConfig::DRPM)
```
diizzyy, line 99: These need to be wrapped around the (optional dependency toggles otherwise CMake will error out because it can't find the target

gahr: good catch!

evan-goode (commented):
Sorry for the delay, LGTM. CI tests pass locally.

> This requires CMake 3.6

Is it 3.7 actually? https://cmake.org/cmake/help/latest/module/FindPkgConfig.html#command:pkg_check_modules says `IMPORTED_TARGET` was added in 3.7.

evan-goode:
Thanks! I also added the minimum CMake version in the spec file.

PR #439, opened 2024-10-01
Include unistd.h for STDOUT_FILENO

gahr:
STDOUT_FILENO is defined in unistd.h, see [POSIX](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/unistd.h.html).

This is required on FreeBSD.

### DIFF
--- a/src/compression_wrapper.c
+++ b/src/compression_wrapper.c
@@ -31,6 +31,7 @@
 #include <lzma.h>
 #ifdef WITH_ZCHUNK
 #include <zck.h>
+#include <unistd.h>
 #endif  // WITH_ZCHUNK
 #include "error.h"
 #include "compression_wrapper.h"


src/compression_wrapper.c
```diff
@@ -26,6 +26,7 @@
 #include <stdio.h>
 #include <ctype.h>
 #include <stdbool.h>
+#include <unistd.h>
```
ppisar, line 29: STDOUT_FILENO is only used if WITH_ZCHUNK macro is defined. Could you please move the #include to an already existing WITH_ZCHUNK block?

gahr: done

ppisar: Great, but please squash the two commits into one.

gahr: done

gahr:
ping - can this be merged?

ppisar (approved):
Thanks. Now it's good.

PR #441, opened 2024-11-07
Allow incremental parsing without filelists or other xml

### LINKED ISSUE #423: Allow parsing packages metadata without filelists
Since Fedora 40, as [filelists are not present in the dnf cache by default](https://fedoraproject.org/wiki/Changes/DNFConditionalFilelists), we should consider making it optional also when parsing XML files in createrepo_c, specifically in `cr_PkgIterator` and the underlying structures.

Several related TODOs are already present in the code in [xml_parser_main_metadata_together.c](https://github.com/rpm-software-management/createrepo_c/blob/1.0.4/src/xml_parser_main_metadata_together.c#L394).

dralley:
Supplying a NULL value as a filelists or other "path" argument ought to work.  But primary metadata is always required.

closes #423

### DIFF
--- a/src/xml_parser_main_metadata_together.c
+++ b/src/xml_parser_main_metadata_together.c
@@ -31,15 +31,16 @@
 #define ERR_DOMAIN      CREATEREPO_C_ERROR
 
 typedef struct {
-    GSList               *in_progress_pkgs_list;
-    int                  in_progress_count_primary;
-    int                  in_progress_count_filelists;
-    int                  in_progress_count_other;
-    GQueue               *finished_pkgs_queue;
-    cr_XmlParserNewPkgCb newpkgcb;      // newpkgcb passed in from user
-    void                 *newpkgcb_data;// newpkgcb data passed in from user
-    cr_XmlParserPkgCb    pkgcb;         // pkgcb passed in from user
-    void                 *pkgcb_data;   // pkgcb data passed in from user
+    GSList                 *in_progress_pkgs_list;
+    int                    in_progress_count_primary;
+    int                    in_progress_count_filelists;
+    int                    in_progress_count_other;
+    GQueue                 *finished_pkgs_queue;
+    cr_XmlParserNewPkgCb   newpkgcb;      // newpkgcb passed in from user
+    void                   *newpkgcb_data;// newpkgcb data passed in from user
+    cr_XmlParserPkgCb      pkgcb;         // pkgcb passed in from user
+    void                   *pkgcb_data;   // pkgcb data passed in from user
+    cr_PackageLoadingFlags loadingflags;  // which callbacks need to be called before the package is considered "finished" loading
 } cr_CbData;
 
 struct _cr_PkgIterator {
@@ -67,17 +68,19 @@ struct _cr_PkgIterator {
 static int
 queue_package_if_finished(cr_Package *pkg, cr_CbData *cb_data)
 {
-    if (pkg && (pkg->loadingflags & CR_PACKAGE_LOADED_PRI) && (pkg->loadingflags & CR_PACKAGE_LOADED_OTH) &&
-        (pkg->loadingflags & CR_PACKAGE_LOADED_FIL))
+    if (pkg && ((pkg->loadingflags & cb_data->loadingflags) == cb_data->loadingflags))
     {
         //remove first element in the list
         cb_data->in_progress_pkgs_list = g_slist_delete_link(cb_data->in_progress_pkgs_list, cb_data->in_progress_pkgs_list);
 
         // One package was fully finished
         cb_data->in_progress_count_primary--;
-        cb_data->in_progress_count_filelists--;
-        cb_data->in_progress_count_other--;
-
+        if (cb_data->loadingflags & CR_PACKAGE_LOADED_FIL) {
+            cb_data->in_progress_count_filelists--;
+        }
+        if (cb_data->loadingflags & CR_PACKAGE_LOADED_OTH) {
+            cb_data->in_progress_count_other--;
+        }
         g_queue_push_tail(cb_data->finished_pkgs_queue, pkg);
     }
     return CR_CB_RET_OK;
@@ -319,8 +322,6 @@ parse_next_section(CR_FILE *target_file, const char *path, cr_ParserData *pd, GE
 
 //TODO(amatej): there is quite some overlap with this and cr_load_xml_files,
 //              consider using this api to implement cr_load_xml_files?
-// TODO: maybe whether or not individual files are parsed could be controlled by NULL paths? I think cr_load_xml_files
-// already works that way.
 cr_PkgIterator *
 cr_PkgIterator_new(const char *primary_path,
                    const char *filelists_path,
@@ -332,8 +333,6 @@ cr_PkgIterator_new(const char *primary_path,
                    GError **err)
 {
     assert(primary_path);
-    assert(filelists_path);
-    assert(other_path);
     assert(!err || *err == NULL);
 
     cr_PkgIterator* new_iter = g_new0(cr_PkgIterator, 1);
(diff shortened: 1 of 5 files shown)


dralley:
This is getting stuck in an infinite loop somewhere.

src/xml_parser_main_metadata_together.c
```diff
@@ -404,9 +416,7 @@ cr_PkgIterator_parse_next(cr_PkgIterator *iter, GError **err) {
     cr_CbData *cbdata = (cr_CbData*) iter->cbdata;
 
     while (!cr_PkgIterator_is_finished(iter) && g_queue_is_empty(cbdata->finished_pkgs_queue)) {
-        while ((cbdata->in_progress_count_primary <= cbdata->in_progress_count_filelists ||
-                cbdata->in_progress_count_primary <= cbdata->in_progress_count_other) &&
-               !iter->primary_is_done)
+        while (((cbdata->in_progress_count_primary <= cbdata->in_progress_count_filelists) && (cbdata->loadingflags & CR_PACKAGE_LOADED_FIL) || (cbdata->in_progress_count_primary <= cbdata->in_progress_count_other) && (cbdata->loadingflags & CR_PACKAGE_LOADED_OTH) || cbdata->loadingflags == CR_PACKAGE_LOADED_PRI) && !iter->primary_is_done)
```
dralley, line 419: Ew.  Suggestions welcome on how to clean this up.

src/xml_parser_main_metadata_together.c
```diff
@@ -370,30 +371,41 @@ cr_PkgIterator_new(const char *primary_path,
...
        return NULL;
     }
-    new_iter->filelists_f = cr_open(filelists_path, CR_CW_MODE_READ, CR_CW_AUTO_DETECT_COMPRESSION, &tmp_err);
-    if (tmp_err) {
-        g_propagate_prefixed_error(err, tmp_err, "Cannot open %s: ", filelists_path);
-        cr_PkgIterator_free(new_iter, err);
-        return NULL;
+    if (new_iter->filelists_path) {
+        new_iter->filelists_f = cr_open(filelists_path, CR_CW_MODE_READ, CR_CW_AUTO_DETECT_COMPRESSION, &tmp_err);
+        cbdata->loadingflags |= CR_PACKAGE_LOADED_FIL;
```
dralley, line 386: If it's too confusing to reuse the same name for the value and the bitfield variants for a slightly different purpose, I can change that.

cbdata->loadingflags == which flags need to be present for the package to be considered "completed"
pkg->loadingflags == which files HAVE been loaded

dralley:
@kontura

src/xml_parser_main_metadata_together.c
kontura: I think that it is fine.

src/xml_parser_main_metadata_together.c
```diff
@@ -67,17 +68,19 @@ struct _cr_PkgIterator {
...
     {
         //remove first element in the list
         cb_data->in_progress_pkgs_list = g_slist_delete_link(cb_data->in_progress_pkgs_list, cb_data->in_progress_pkgs_list);
 
         // One package was fully finished
         cb_data->in_progress_count_primary--;
-        cb_data->in_progress_count_filelists--;
-        cb_data->in_progress_count_other--;
-
+        if (cb_data->loadingflags | CR_PACKAGE_LOADED_FIL) {
```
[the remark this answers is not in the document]
dralley: Yes

src/xml_parser_main_metadata_together.c
```diff
@@ -319,8 +322,6 @@ parse_next_section(CR_FILE *target_file, const char *path, cr_ParserData *pd, GE
 
 //TODO(amatej): there is quite some overlap with this and cr_load_xml_files,
 //              consider using this api to implement cr_load_xml_files?
-// TODO: maybe whether or not individual files are parsed could be controlled by NULL paths? I think cr_load_xml_files
-// already works that way.
 cr_PkgIterator *
 cr_PkgIterator_new(const char *primary_path,
                    const char *filelists_path,
```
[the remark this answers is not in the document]
dralley: Yes, but now I'm confused as to why the tests even work, because they should be exercising the path where these values are NULL.

tests/python/tests/test_xml_parser.py
```diff
@@ -1245,6 +1245,46 @@ def warningcb(warn_type, msg):
...
+        self.assertEqual(packages[0].name, "fake_bash")
+        self.assertEqual(packages[0].changelogs, [
+            ('Tomas Mlcoch <tmlcoch@redhat.com> - 1.1.1-1', 1334664000, '- First release'),
+        ])
+        self.assertEqual(packages[0].files, [(None, '/usr/bin/', 'fake_bash')])  # the files still get parsed - only from primary.xml
+        self.assertListEqual(warnings, [])
+        self.assertRaises(StopIteration, next, package_iterator)
+        self.assertTrue(package_iterator.is_finished())
+
+    def test_xml_parser_pkg_iterator_repo02_no_other(self):
```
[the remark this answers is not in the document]
dralley: Testing that it fails you mean?

src/xml_parser_main_metadata_together.c
dralley: I can take another look at it, but I had trouble getting the conditionals such that it wouldn't infinite loop or terminate too early.

dralley: The tests seem to pass but I worry that this isn't going to work if the file doesn't fit completely inside the first XML buffer.  The conditional is "&&" so when it comes back around it's never going to move forwards if the filelists or other in progress haven't moved from 0.

dralley: Well, I did a test with a smaller buffer and it still seems to work so I guess I'm missing something.

src/xml_parser_main_metadata_together.c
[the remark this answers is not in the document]
kontura: It is because the `asserts` work only in `debug` build. 
It does fail for me locally.

src/xml_parser_main_metadata_together.c
kontura: Some problems with infinite looping or terminating too early were fixed by the `|` -> `&` switch up.

> The tests seem to pass but I worry that this isn't going to work if the file doesn't fit completely inside the first XML buffer. The conditional is `&&`, so when it comes back around after the first buffer gets processed it's never going to move forwards because `in_progress_primary_count > 0` and `in_progress_{filelists|other} == 0`.

No, the `in_progress_primary_count` doesn't count packages that have not finished parsing primary. It counts packages that have fully parsed primary but are still in progress (require filelists/other to be parsed). The same goes for `in_progress_filelists_count` and `in_progress_other_count`.
I now see that the naming is confusing. :disappointed:

kontura: > Though both the proposed version and the unchanged version will parse the whole file at once if just `primary` is provided (loading it all to memory at once). 🤔

I was thinking how to make this work and one solution I came up with is to change:
```diff
-        while ((cbdata->in_progress_count_primary <= cbdata->in_progress_count_filelists ||
-                cbdata->in_progress_count_primary <= cbdata->in_progress_count_other) &&
-                !iter->primary_is_done)
-        {
+        if (!iter->primary_is_done) {
```

This makes sense to me because `primary` is mandatory so the main loop parses `primary` and the nested `while` loops just make sure `filelists` and `other` catch up if needed.

To further simplify it I think we could also remove the comparisons of `in_progress_count_other` and `in_progress_count_filelists` like:
```diff
             }
         }

-        while ((cbdata->in_progress_count_filelists <= cbdata->in_progress_count_primary ||
-                cbdata->in_progress_count_filelists <= cbdata->in_progress_count_other) &&
-               !iter->filelists_is_done)
-        {
+        while ((cbdata->in_progress_count_filelists <= cbdata->in_progress_count_primary) && !iter->filelists_is_done) {
             iter->filelists_is_done = parse_next_section(iter->filelists_f, iter->filelists_path, iter->filelists_pd, err);
             if (*err) {
                 return NULL;
             }
         }

-        while ((cbdata->in_progress_count_other <= cbdata->in_progress_count_filelists ||
-                cbdata->in_progress_count_other <= cbdata->in_progress_count_primary) &&
-               !iter->other_is_done)
-        {
+        while ((cbdata->in_progress_count_other <= cbdata->in_progress_count_primary) && !iter->other_is_done) {
             iter->other_is_done = parse_next_section(iter->other_f, iter->other_path, iter->other_pd, err);
             if (*err) {
                 return NULL;
```

This changes ordering of warnings in a couple of tests but those can be amended.

It also makes this whole `while` madness a bit simpler.

dralley: Seems like it does more than change the ordering?

src/xml_parser_main_metadata_together.c
```diff
@@ -404,29 +414,23 @@ cr_PkgIterator_parse_next(cr_PkgIterator *iter, GError **err) {
     cr_CbData *cbdata = (cr_CbData*) iter->cbdata;
 
     while (!cr_PkgIterator_is_finished(iter) && g_queue_is_empty(cbdata->finished_pkgs_queue)) {
-        while ((cbdata->in_progress_count_primary <= cbdata->in_progress_count_filelists ||
-                cbdata->in_progress_count_primary <= cbdata->in_progress_count_other) &&
-               !iter->primary_is_done)
+        while (!iter->primary_is_done)
```
kontura, line 417: This has to be: `if (!iter->primary_is_done)`.

dralley: Ah, whoops.

src/xml_parser_main_metadata_together.c
kontura: Ok, now there is no reordering, it seems the buffer size change (which was in your temp commit) is also needed for it to happen.

kontura (approved):
Thank you!

PR #446, opened 2025-03-16
Remove LEGACY_HASHES override from setup.py

dralley:
Pulp needed this overridden at one point, but not anymore.


dralley:
A release with this would be *nice* but it's not too time sensitive for us.

PR #466, opened 2026-02-18
Remove uneeded inclusions of <rpm/rpmlib.h>

ppisar:
The header file is not actually needed to be included on matny places, especially in public API createrepo_c header files as given files do not use any RPM symbols.


m-blaha (approved):
Thank you!

PR #464, opened 2026-02-13
Avoid stat as struct member name to prevent macro conflicts on AIX

Lakshmi-Surekha:
Hi All,

While building createrepo_c on AIX with large file support, we encountered a compilation issue caused by a name collision between a struct member and a system macro.

In threads.h, the cr_CompressionTask struct currently has a member:

`cr_ContentStat *stat;`

On AIX, D_LARGE_FILES defines the macro:

`#define stat stat64`

During compilation, any use of task->stat in the code gets preprocessed to task->stat64, which does not exist in the struct, resulting in errors like:

```
[ 25%] Building C object src/CMakeFiles/libcreaterepo_c.dir/threads.c.o
cd /opt/freeware/src/packages/BUILD/createrepo_c-1.2.1/build-py3/src && /opt/freeware/bin/gcc -DENABLE_LEGACY_WEAKDEPS=1 -DG_LOG_DOMAIN=\"C_CREATEREPOLIB\" -Dlibcreaterepo_c_EXPORTS -I/opt/freeware/src/packages/BUILD/createrepo_c-1.2.1 -I/opt/freeware/include/glib-2.0 -I/opt/freeware/lib/glib-2.0/include -I/opt/freeware/include/libxml2 -fPIC -fpermissive -D_ALL_SOURCE -D_LARGE_FILES -D_XOPEN_SOURCE=700 -static-libgcc -static-libstdc++ -std=gnu99 -O2 -g -DNDEBUG -fPIC -MD -MT src/CMakeFiles/libcreaterepo_c.dir/threads.c.o -MF CMakeFiles/libcreaterepo_c.dir/threads.c.o.d -o CMakeFiles/libcreaterepo_c.dir/threads.c.o -c /opt/freeware/src/packages/BUILD/createrepo_c-1.2.1/src/threads.c
cc1: warning: command-line option '-fpermissive' is valid for C++/ObjC++ but not for C
In file included from /usr/include/rpm/rpmio.h:11,
                 from /usr/include/rpm/rpmlib.h:11,
                 from /opt/freeware/src/packages/BUILD/createrepo_c-1.2.1/src/dumper_thread.h:28,
                 from /opt/freeware/src/packages/BUILD/createrepo_c-1.2.1/src/threads.c:24:
/opt/freeware/src/packages/BUILD/createrepo_c-1.2.1/src/threads.c: In function 'cr_compressiontask_new':
[...]

PR #469, opened 2026-02-23
Restore compatibility with RPM 4.14.3

ppisar:
Commit 9ff916d5fa738c73bd9ee796d67d29c65b4b7dfb ("Remove uneeded inclusions of <rpm/rpmlib.h>") broke comptability with RPM before 4.16.0-beta2. Building with 4.14.3 failed:

    /tmp/createrepo_c/src/misc.c:28:10: fatal error: rpm/rpmver.h: No such file or directory
     #include <rpm/rpmver.h>
              ^~~~~~~~~~~~~~
This patch restores a compatibility with older rpm versions. Tested with rpm-4.14.3.

(In the future, createrepo_c should define a minimal supported RPM version.)

Relates: #464

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -78,6 +78,11 @@ IF (NOT HAVE_NFTW)
     ENDIF ()
 ENDIF ()
 
+# rpm/rpmver.h exists since RPM 4.16.0-beta2.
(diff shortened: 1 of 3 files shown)


m-blaha (approved):
Thank you!

PR #470, opened 2026-02-25
Package from rpm file digests

kontura:
Follow up to https://github.com/rpm-software-management/createrepo_c/pull/362

Add `file_digests` flag to restore old Python API to get files represented as a tuple with 3 values.

Another option could be to expose `cr_HeaderReadingFlags flags` directly.

### DIFF
--- a/src/parsehdr.h
+++ b/src/parsehdr.h
@@ -39,6 +39,11 @@ typedef enum {
     CR_HDRR_NONE            = (1 << 0),
     CR_HDRR_LOADHDRID       = (1 << 1), /*!< Load hdrid */
     CR_HDRR_LOADSIGNATURES  = (1 << 2), /*!< Load siggpg and siggpg */
+    CR_HDRR_NOFILEDIGESTS   = (1 << 3), /*!< Don't load file digests */
+    CR_HDRR_ALL             = CR_HDRR_NONE |
+                              CR_HDRR_LOADHDRID |
+                              CR_HDRR_LOADSIGNATURES |
+                              CR_HDRR_NOFILEDIGESTS, /*!< All flags set (doesn't mean load all) */
 } cr_HeaderReadingFlags;
 
 /** Read data from header and return filled cr_Package structure.
--- a/src/python/createrepo_c/__init__.py
+++ b/src/python/createrepo_c/__init__.py
@@ -86,6 +86,11 @@ XMLFILE_OTHER         = _createrepo_c.XMLFILE_OTHER         #: Other xml file
 XMLFILE_PRESTODELTA   = _createrepo_c.XMLFILE_PRESTODELTA   #: Prestodelta xml file
 XMLFILE_UPDATEINFO    = _createrepo_c.XMLFILE_UPDATEINFO    #: Updateinfo xml file
 
+HDRR_NONE           = _createrepo_c.HDRR_NONE           #: No header reading flags
+HDRR_LOADHDRID      = _createrepo_c.HDRR_LOADHDRID      #: Load hdrid
+HDRR_LOADSIGNATURES = _createrepo_c.HDRR_LOADSIGNATURES #: Load siggpg and sigpgp
+HDRR_NOFILEDIGESTS  = _createrepo_c.HDRR_NOFILEDIGESTS  #: Don't load file digests (changes package file tuples from 4 to 3 elements)
+
 #: XML warning - Unknown tag
 XML_WARNING_UNKNOWNTAG  = _createrepo_c.XML_WARNING_UNKNOWNTAG
 
--- a/src/python/createrepo_c/__init__.py
+++ b/src/python/createrepo_c/__init__.py
@@ -314,10 +314,10 @@ class UpdateInfoXmlFile(XmlFile):
 # Functions
 
 def package_from_rpm(filename, checksum_type=SHA256, location_href=None,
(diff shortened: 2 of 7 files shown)


dralley (approved):
Thank you, this is exactly what I was thinking of. 👍

src/parsehdr.h
```diff
@@ -39,6 +39,7 @@ typedef enum {
     CR_HDRR_NONE            = (1 << 0),
     CR_HDRR_LOADHDRID       = (1 << 1), /*!< Load hdrid */
     CR_HDRR_LOADSIGNATURES  = (1 << 2), /*!< Load siggpg and siggpg */
+    CR_HDRR_NOFILEDIGESTS   = (1 << 3), /*!< Load file digests */
```
dralley: Why the naming convention disparity between `nofiledigests` and `loadsignatures`?  Mostly to keep the options off by default?

Unrelated: there is not actually a way to access package signatures via Python yet, is there?  Would you be averse to having one?

kontura: > Why the naming convention disparity between `nofiledigests` and `loadsignatures`? Mostly to keep the options off by default?
> 

Yes, it is to preserve the current default behavior of API like  `cr_package_from_header`. I would also prefer if we could do `CR_HDRR_LOADFILEDIGESTS` but unfortunately that would be a breaking change now.

> Unrelated: there is not actually a way to access package signatures via Python yet, is there? Would you be averse to having one?

Indeed there is not and I would be fine with adding it. We cannot even configure these flags from Python (there [is a TODO](https://github.com/rpm-software-management/createrepo_c/blob/master/src/python/parsepkg-py.c#L39) for this).
I was thinking we could expose the flags directly (since we want to do that anyway) instead of the specific `file_digests` option. To get the old 3 elements file tuple the user would then specify `CR_HDRR_NOFILEDIGESTS` manually.
Yesterday I chose to use separate `file_digests` option only because it somehow felt more appropriate for this kind of "hotfix" but thinking about it now I guess we might as well do it now and skip the `file_digests`?

dralley: I would be fine with either solution, as long as it's accessible

dralley (approved):

dralley:
@kontura If this is ready to merge, I have a followup PR for the signatures that requires the flags feature

kontura:
> @kontura If this is ready to merge, I have a followup PR for the signatures that requires the flags feature

It is, I am just waiting for someone else from the team to review it.

src/python/parsepkg-py.c
```diff
@@ -36,20 +36,27 @@ py_package_from_rpm(G_GNUC_UNUSED PyObject *self, PyObject *args)
...
                                          &location_href,
                                          &location_base,
-                                         &changelog_limit)) {
+                                         &changelog_limit,
+                                         &header_reading_flags)) {
+        return NULL;
+    }
+
+    if (header_reading_flags & ~ CR_HDRR_ALL) {
+        nice_exception(&tmp_err, "Unknown header reading flags.");
```
evan-goode: More importantly, `tmp_err` is NULL here causing segfault.

src/python/createrepo_c/__init__.py
```diff
@@ -86,6 +86,11 @@
 XMLFILE_PRESTODELTA   = _createrepo_c.XMLFILE_PRESTODELTA   #: Prestodelta xml file
 XMLFILE_UPDATEINFO    = _createrepo_c.XMLFILE_UPDATEINFO    #: Updateinfo xml file
 
+HDRR_NONE           = _createrepo_c.HDRR_NONE           #: No header reading flags
+HDRR_LOADHDRID      = _createrepo_c.HDRR_LOADHDRID      #: Load hdrid
+HDRR_LOADSIGNATURES = _createrepo_c.HDRR_LOADSIGNATURES #: Load siggpg and sigpgp
+HDRR_NOFILEDIGESTS  = _createrepo_c.HDRR_NOFILEDIGESTS  #: Don't load file digests (changes package file tuples from 4 to 3 elements)
+
```
evan-goode, line 93: Could be useful to have `HDRR_ALL` here too.

evan-goode (changes requested):
LGTM otherwise!

src/python/parsepkg-py.c
kontura: Good catch, thank you!

src/python/createrepo_c/__init__.py
kontura: I was thinking about it but I am not that happy with the `HDRR_ALL` because the original two flags `CR_HDRR_LOADHDRID` and `CR_HDRR_LOADSIGNATURES` add stuff but the new one `CR_HDRR_NOFILEDIGESTS` subtracts. This is to preserve API behavior as mentioned above. 
I worry someone could use the `HDRR_ALL` wrongly thinking they will get everything possible so I would rather not advertise it further.

On the other hand it would be documented so I have no problem exposing it in Python if you prefer.

evan-goode: Yeah, I was thinking of a Python consumer validating flags similar to `header_reading_flags & ~ CR_HDRR_ALL`. But that's niche and you're right, it could be easily confused.

evan-goode (approved):

dralley:
@kontura Can we have an upstream release with this?

PR #477, opened 2026-03-06
Add (non-legacy) signature access via Python

dralley:
This is an addition to the C API as well, strictly speaking.

Assisted By: claude sonnet 4.6


src/parsehdr.c
```diff
@@ -702,6 +709,42 @@ cr_package_from_header(Header hdr,
 
         rpmtdFree(gpgtd);
         rpmtdFree(pgptd);
+
+        rpmtd openpgptd = rpmtdNew();
+
+        // https://github.com/rpm-software-management/rpm/issues/4137
+        if (headerGet(hdr, RPMTAG_OPENPGP, openpgptd, HEADERGET_MINMEM)
```
dralley, line 716: Does RPM just ignore unknown tags entirely?  I've hardcoded the value, but it does not seem to help - on OS' that don't have the RPMTAG_OPENPGP patches it doesn't seem like this codepath ever gets taken (and therefore only one signature is returned on the multi-signature package).  Does it get filtered out of the `Header`?

dralley: Basically it seems that on new versions of Fedora the tests correctly see 2 signatures, and on older versions they do not, and only return the one V4 signature (either RSAHEADER or DSAHEADER)

pmatilai: It's complicated :sweat_smile: 

For a normal tag, you would get the data based on its number regardless of whether rpm knows about it or not. But RPMTAG_OPENPGP resides in the signature header originally, and during rpmReadPackageFile() only select tags are copied into the header that is returned. The exact behavior varies between rpm versions, but all recent versions only copy a set of tags they know about and so upstream rpm 4.x will not see RPMTAG_OPENPGP at all.

dralley: Thanks for the explanation!

Not sure what to do about the tests then - perhaps just ignore them until all platforms in the CI know about RPMTAG_OPENPGP?

src/package.h
```diff
@@ -128,10 +128,11 @@ typedef struct {
                                      cr_PackageFile structs) */
     GSList *changelogs;         /*!< changelogs (list of cr_ChangelogEntry
                                      structs) */
+    GSList *signatures;         /*!< signatures (list of char*)*/
```
kontura, line 131: Adding to a public struct like this breaks the C library ABI.
I know we did it in the past but I know better now.

We will need to bump the major version and it might make sense to introduce some p_impl pattern here. It seems this struct needs to be extended often.

Also can I ask why you chose to use `char*` instead of `cr_BinaryData`?

src/parsehdr.c
```diff
@@ -45,6 +45,11 @@
 #define RPMSENSE_MISSINGOK (1 << 19)
 #endif
 
+#ifndef RPMTAG_OPENPGP
+// hardcode RPMTAG_OPENPGP value, because we can't rely on it being defined, and we don't want to miss signatures
+#define RPMTAG_OPENPGP (RPMTAG_SIG_BASE + 22)
```
kontura, line 50: If I understood @pmatilai explanation correctly this won't help right? If rpm doesn't know about the tag it won't copy it into the returned header.

Would it make sense to drop this hardcoding and use the `#ifndef RPMTAG_OPENPGP` to determine whether to run the unittest with v6 signatures?

dralley: Well, while I would have hoped it would work as expected, it still avoids needing to make the build more complicated.  This will build everywhere and work without needing to be rebuilt once RPM gets updated on those platforms.

That strategy works for the C tests but doesn't work for the Python tests

With that said I can still see it both ways, so maybe it makes sense to drop.

src/package.h
dralley: The signatures stored in RPMTAG_OPENPGP are natively base64-encoded strings unlike the legacy signatures.    I figured it would make sense to return them in their native form.  Also we can store them directly in the chunk storage this way.

dralley:
@kontura We had discussed that this might need to go into the next major release to avoid breaking ABI

What is the intended timeline on that?  Would it be realistic to get that done within the next 2-3 months?

kontura:
> What is the intended timeline on that? Would it be realistic to get that done within the next 2-3 months?

Unfortunately I am not sure we will have time to do everything I would like for the next major bump: [list link](https://github.com/rpm-software-management/createrepo_c/issues?q=state%3Aopen%20label%3A%22breaking%20change%22).

We probably could do a major bump but with only a couple of the changes (the more important ones). 

If we managed to do https://github.com/rpm-software-management/createrepo_c/issues/483 we could get this PR in and it wouldn't even be a break.

dralley:
This is somewhat less time sensitive now that we've discovered that RPM on RHEL 9 cannot even access the SIGTAG_OPENPGP tag, admittedly.  We don't want to deal with platform-dependent behavior.

As a complete aside, I discovered the other day that you tell RPM (on RHEL 9) to sign a package using a standard OpenPGPv4 key generated by Sequoia using a common algorithm like RSA, it'll do it, but it cannot necessarily verify that same package signature because when importing the public key, the built-in packet parsing doesn't seem to handle non-mainkey signing keys (and maybe some other modern flags).

pmatilai:
RHEL 9 RPM does support subkeys, but I seem to recall there being some other incompatibilities. RHEL 9 RPM was only ever tested with GnuPG-signing. There basically never was any guarantee RPM can verify what was signed because the signing and verification are done by different programs from different origins. RPM checks for some basics like supported signature algorithm, but there are any number of other stuff...

dralley:
I forget the exact specifics, you might be correct that it was mostly an issue with supported flags rather than the subkeys specifically.  In any case, we're going to use my own RPM header parsing code for at least as long as it takes until we can drop EL9 from our support matrix entirely.

src/dumper_thread.c
```diff
@@ -356,10 +356,6 @@ get_checksum(const char *filename,
         cr_ChecksumCtx *ctx = cr_checksum_new(type, err);
         if (!ctx) return NULL;
 
-        if (pkg->siggpg)
-            cr_checksum_update(ctx, pkg->siggpg->data, pkg->siggpg->size, NULL);
-        if (pkg->sigpgp)
-            cr_checksum_update(ctx, pkg->sigpgp->data, pkg->sigpgp->size, NULL);
         if (pkg->hdrid)
```
dralley, line 370: TODO: probably need to checksum package-> signatures, yes?

src/package_internal.h
```diff
@@ -78,8 +78,8 @@ struct _cr_Package {
                                      structs) */
 
     char *hdrid;
-    cr_BinaryData *siggpg;
-    cr_BinaryData *sigpgp;
+
+    GSList *signatures;         /*!< signatures (list of char*)*/
```
dralley, line 84: TODO: confirm that returning base64 encoded strings makes sense.  It's what SIGTAG_OPENPGP does.  But, in theory, we could just return the raw signature bytes as cr_BinaryData.

src/package_internal.h
```diff
@@ -78,8 +78,8 @@ struct _cr_Package {
                                      structs) */
 
     char *hdrid;
-    cr_BinaryData *siggpg;
-    cr_BinaryData *sigpgp;
```
dralley, line 82: @kontura Would you be comfortable with removing these, or would you rather keep them for 2.0?

kontura: Personally I would go with just a deprecation for now.

dralley: Note that we've just removed access to the fields, and the getter functions are brand new (in https://github.com/rpm-software-management/createrepo_c/commit/fbe4c23d47ea3954e8f6ae8616044a51985d0870#diff-677d49b47b986358bab572cb65a2759d17862742c65d338b52428e3b4686e100R163).  So 2.0 will have the pgpsig / sigpgp getter methods be deprecated from the very release they are added, while the old API is removed entirely in a breaking change.

If you still want to keep them though, that's fine.

src/package_internal.h
dralley: I think it makes sense.  Better than allocating and freeing many related objects.

dralley:
@kontura I'm going on vacation for the next week and a half starting Thursday afternoon

re: the deprecation vs. removal, I pushed both approaches, I can squash it to whatever we want tomorrow morning and hopefully merge this before I go on leave.  I added the cache checksumming code for `pkg->signatures`.

When I get back I will have a few more patches to finish the Abi3 issue and a few other assorted things.

src/package_internal.h
kontura: Yea I know but while the API changes I think we should preserve the functionality (data from `RPMTAG_SIGGPG` and `RPMTAG_SIGPGP` is still accessible).

dralley:
@kontura Re-pushed.

src/package.h
```diff
@@ -715,28 +715,28 @@ void cr_package_set_changelogs(cr_Package *package, GSList *list);
...
  */
 cr_BinaryData *cr_package_get_siggpg(cr_Package *package);
 
-/** Set OpenPGP signature.
+/** DEPRECATED: Set OpenPGP signature.
  * Ownership: data is transferred to the package.
  * @param package       cr_Package
  * @param data          OpenPGP signature
  */
 void cr_package_set_siggpg(cr_Package *package, cr_BinaryData *data);
```
dralley, line 730: I would recommend not adding this API in the first place, to be honest. It serves no purpose, `createrepo_c` cannot do anything with the signature other than hold it since it has no impact on the metadata.  The value comes from parsing / extracting the signatures, not from being able to set it on the handle.

kontura: That is a very good point, I will address that in a follow up PR.

kontura (approved):
Thank you for all the work!

PR #478, opened 2026-03-20
Fix GHA Python builds, again

dralley:
Hopefully for the last time in a while.

* Python 3.14 "free threaded" builds were broken - 1 line fix
* Replace scikit-build with scikit-build core
* Replace VERSION.cmake with a VERSION file that is easily parsed by both the Python builder and Cmake
* Get rid of setup_for_python_metadata.py


kontura (approved):
Thanks!

Issue #424, opened 2024-03-01
cr_xml_dump_int() should point to a forbidden character


ppisar:
struct cr_XmlStruct cr_xml_dump_int(cr_Package *pkg, gboolean filelists_ext, GError **err) correctly rejects a package whose metadata cannot be represented in XML:

~~~~
    if (cr_Package_contains_forbidden_control_chars(pkg)) {
        g_set_error(err, CREATEREPO_C_ERROR, CRE_XMLDATA,
                    "Forbidden control chars found (ASCII values <32 except 9, 10 and 13).");
        return result;
    }
~~~~

But the callers would like to know where the forbidden character is. Requested in <https://github.com/rpm-software-management/createrepo_c/pull/165#issuecomment-1973088771>.

Because cr_Package_contains_forbidden_control_chars() and lower functions are in public createrepo_c/xml_dump.h header file and they miss an argument for returning the precise location, a fix will need to add new functions (and wrap the current ones) not to break ABI.

sergiomb2:
reference https://github.com/fedora-copr/copr/issues/3163

ppisar:
PR #425 added printing warning messages which pinpoints a package name, RPM tag name and RPM tag value.

This issue is kept open to:

* Report exact byte offset in the tag values (it's difficult to locate the offending nonprintablecharacters in the output).
* Store the warnings into an error buffer and let the caller to print it.

PR #479, opened 2026-03-24
Strip control characters on package header parse

### LINKED ISSUE #327: Critical: Cannot dump XML on fedora package
This package: python-pyABF-2.3.6-2.fc38 ( https://koji.fedoraproject.org/koji/buildinfo?buildID=2067454 ) 

Causes createrepo_c to output: 

C_CREATEREPOLIB: Critical: Cannot dump XML for python3-pyABF (3ab9a1aafc8cc037c279fad85f5d34726a4491f763871b80df494aa9f2856e14): Forbidden control chars found (ASCII values <32 except 9, 10 and 13).
17:15:51: CACHE HIT python3-pyasn1-modules-0.4.8-11.fc37.noarch.rpm

The only thing that looks at all out of the ordinary is a changelog entry: 

- added patch [200~9cd568066b5ff8d6079e2d8cf8af388066cc9b79~

dralley:
The simplest place to add this is in cr_safe_string_chunk_insert()

Do not fatally error when trying to dump them.

closes #327
Assisted-By: claude-opus-4.6


dralley:
Upside to this approach: 

* Simple
* Universal - works on setters (via Python) and also parsing direct from header and also copying between package objects.

Downside to this approach: 

* It would be difficult to implement https://github.com/rpm-software-management/createrepo_c/issues/424 with this approach because you don't know exactly which field you're working with
* Maybe duplicative.  `cr_package_copy` could be duplicating work already done - inefficient

src/xml_dump.h
```diff
@@ -212,22 +212,6 @@ gboolean cr_hascontrollchars(const unsigned char *str);
...
- */
-gboolean cr_Package_contains_forbidden_control_chars(cr_Package *pkg);
-
-/** Check if list of cr_Dependency stucts contains any strings with chars
- * with value <32 (except 9, 10 and 13), using cr_hascontrollchars
- *
- * @param deps          the GSList of cr_Dependencies in question
- * @return              boolean value
- */
-gboolean cr_GSList_of_cr_Dependency_contains_forbidden_control_chars(GSList *deps);
```
kontura, line 229: We cannot drop these without breaking API (it would require major version bump).

kontura (commented):
I like how simple and elegant this is however in this case I would personally rather have the ugly code with the ability to point the user closer to the problem.

Plus I was thinking we should relax the behavior only for certain fields. I can't image a package with control characters in dependencies is ever wanted.

src/xml_dump.h
dralley: So just leave them as dead code then?

dralley: Or - can we do the version bump in the next month or two?

kontura: Either way I would be fine with leaving it in unused with a deprecation notice.

Though I am still not convinced of the approach currently used in the PR.

PR #480, opened 2026-03-24
Add a bounds check on dirlist indexing

dralley:
Protect against cases where `dir_list[dir_idx]` might go out-of-bounds.

Assisted-By: claude-opus-4.6

### DIFF
--- a/src/parsehdr.c
+++ b/src/parsehdr.c
@@ -281,10 +281,10 @@ cr_package_from_header(Header hdr,
 
     // Create list of pointer to directory names
 
-    int dir_count;
+    uint32_t dir_count;
     char **dir_list = NULL;
     if (headerGet(hdr, RPMTAG_DIRNAMES, dirnames,  flags) && (dir_count = rpmtdCount(dirnames))) {
-        int x = 0;
+        uint32_t x = 0;
         dir_list = malloc(sizeof(char *) * dir_count);
         while (rpmtdNext(dirnames) != -1) {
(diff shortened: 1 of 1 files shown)


src/parsehdr.c
```diff
@@ -316,7 +316,12 @@ cr_package_from_header(Header hdr,
             cr_PackageFile *packagefile = cr_package_file_new();
             packagefile->name = cr_safe_string_chunk_insert(pkg->chunk,
                                                          rpmtdGetString(filenames));
-            packagefile->path = (dir_list) ? dir_list[(int) rpmtdGetNumber(indexes)] : "";
+            int dir_idx = (int) rpmtdGetNumber(indexes);
```
ppisar: If you don't trust what librpm returns and you validate the index value, then please also check that the unint64_t value returned from rpmtdGetNumber() fits into int type.

dralley: Done.  Except no need to check, we can just cast to the 64 bit types.

dralley: It's less that librpm isn't trusted, so much as the RPMs can't be 100% trusted, so it can't hurt to be a bit more defensive

ppisar: That's also a possible fix. But then you compare signet int and uint32_t in `assert(x == dir_count);` at line 293. Please also change the type of `x`.

Maybe using  `rpm_count_t` instead of `uint32_t` could make the choice of the type self-documenting.

ppisar (approved):
It's good now. Thanks for the patch.

Issue #484, opened 2026-05-01
Build Python packages against the "abi3" limited / stable API, so we don't need separate packages per each Python version


dralley:
As it stands, every time a new version of Python is released it forces an update of the CI (specifically `cibuildwheel`) to build packages that support the new versions, and of course a new release as the existing releases were not built for the new versions.

There is a "limited" Python API, which if you build against it, it enables building packages which work against multiple Python versions (stable-ish ABI).  Unfortunately we cannot currently build against it because several off-limits features are used by the python extension code.

Most of these changes are considered good hygiene, anyway, and ought to assist with the opaque pointer changes.

(list generated by Claude Opus 4.6)

---
## Main Blockers
1. 16 static PyTypeObject definitions (HARD)
Every type (Package, Repomd, UpdateRecord, etc.) is a statically-defined PyTypeObject initialized with PyType_Ready(). Under the limited API, PyTypeObject is opaque — these all need converting to `PyType_FromSpec() + PyType_Slot[]` heap types, with type objects stored in module state instead of globals.

2. 16 *Object_Check macros referencing global type addresses (HARD)
Each header defines e.g. `PackageObject_Check(o) as PyObject_TypeCheck(o, &Package_Type)`. With heap types, these need module-state-aware lookups instead of 1&Package_Type1. Tightly coupled with 1.

3. 1Py_TYPE(self)->tp_free(self)1 in all 16 dealloc functions (EASY)
Accesses tp_free through the opaque PyTypeObject. Replace with `PyType_GetSlot(Py_TYPE(self), Py_tp_free)` (available since 3.9).

4. Direct ob_refcnt access in xml_parser-py.c:84 (MODERATE)
`result->ob_refcnt == 1` — PyObject is opaque under limited API. `Replace with Py_REFCNT()` (function form in 3.10+), though the ownership-check pattern itself is fragile.

5. Single-phase module init (MODERATE)
Currently uses `PyModule_Create()`. Multi-phase init (Py_mod_exec slots, PEP 489) is needed to properly support per-module state for the heap types.

## Not Blocking but Worth Modernizing

`PyModule_AddObject (17 calls) → PyModule_AddType() / PyModule_AddObjectRef() (3.10+)`

## Bottom Line

[...]

PR #485, opened 2026-05-05
Initial Python module modernization steps

dralley:
re #484

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,6 @@ classifiers = [
     "Topic :: System :: Software Distribution",
     "Topic :: System :: Systems Administration",
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.9",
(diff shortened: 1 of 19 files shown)


dralley:
Not sure why the Python workflows seem to need approval?

I suppose it's failing on EPEL 9 because of the Python 3.10 requirement (> 3.9).  Is 3.9 support a hard requirement for new versions?

RHEL 9 does actually have Python 3.11 in appstream, it's just not the default.  3.9 goes EOL (upstream) in a couple of months.

Satellite uses Python 3.11, so while we support RHEL 9 this would not cause issues for us.

kontura:
> I suppose it's failing on EPEL 9 because of the Python 3.10 requirement (> 3.9). Is 3.9 support a hard requirement for new versions?

I think it is fine.

Thank you for the PR!

kontura (approved):

PR #486, opened 2026-05-06
Python extension modernization / abi stability / bugfixes - part 2

dralley:


src/python/updaterecord-py.c
```diff
@@ -290,7 +287,6 @@ get_datetime(_UpdateRecordObject *self, void *member_offset)
         res = strptime(str, "%Y-%m-%d", dt);
         if (res == NULL) {
             g_free(dt);
-            // Try to convert the whole string to a number if it passes it's likely in epoch format
```
kontura, line 293: This commit is dropping 3 comments is that intentional? I would prefer to keep them.

src/python/updaterecord-py.c
```diff
@@ -399,17 +412,31 @@ set_datetime(_UpdateRecordObject *self, PyObject *value, void *member_offset)
...
+    int second = _get_int_attr(value, "second");
+    if (PyErr_Occurred()) return -1;
+
     char *date = malloc(20 * sizeof(char));
     snprintf(date, 20, "%04d-%02d-%02d %02d:%02d:%02d",
-             PyDateTime_GET_YEAR(value) % 9999, PyDateTime_GET_MONTH(value) % 13,
-             PyDateTime_GET_DAY(value) % 32, PyDateTime_DATE_GET_HOUR(value) % 24,
-             (PyDateTime_DATE_GET_MINUTE(value) % 60), PyDateTime_DATE_GET_SECOND(value) % 60);
+             year % 9999, month % 13, day % 32,
+             hour % 24, minute % 60, second % 60);
```
kontura: This now produces a compile time warnings.
I think the problem is that if `value` is negative the modulo will also be negative and that would require more space in the buffer. Could we perhaps get the values (year, month, day, ...) as unsigned?

kontura (changes requested):

src/python/updaterecord-py.c
dralley: Would you prefer using unsigned, or just increasing the buffer size slightly (e.g. to 32)?

kontura: I probably find unsigned a bit cleaner implementation wise but I am fine either way. I think python stores only a byte or two for each value so it shouldn't happen and if it does the values will be nonsensical in both approaches.

dralley: Updated

src/python/updaterecord-py.c
kontura: There are still 2 comments being removed: `/* Length 13 is plenty of space for epoch. */` and `// Try to convert the whole string to a number if it passes it's likely in epoch format`.


Otherwise it looks good.

dralley: I think I did `git restore --staged -p` and then accidentally re-added the deletions when committing.

Anyway, fixed.

kontura (approved):
Thank you!

dralley:
Thanks!

Issue #483, opened 2026-04-27
Use opaque pointers for public structs


kontura:
Using opaque pointers for public structs would allow us to change them without breaking ABI.

We should investigate where would this pattern make sense and if it is possible on such places.

Particularly `cr_Package` would be a good candidate as it seems to be changed a lot [[1](https://github.com/rpm-software-management/createrepo_c/pull/477)] but we should look into the rest of the API as well.

kontura:
Resolved by: https://github.com/rpm-software-management/createrepo_c/pull/489

PR #487, opened 2026-05-07
Rename "requires" field of cr_Package structure to "requirements"

### LINKED ISSUE #181: Use of reserved C++ keyword
In package.h the package struct contains the line `GSList *requires`. Since C++20 "requires" is a reserved keyword, so clang++ is complaining about it and terminates compilation of my code. I am using clang version 8.0.0 with the `--std=c++2a` flag. Using g++ does not lead to errors (version 9.1.0).

Since this is code that is accessible to clients of createrepo_c, changing the variable's name is probably not much of an option because of the amount of code that might break. Any tips how I can solve this? Unfortunately, using an older C++ standard is not an option.

ppisar:
C++20 added "requires" keyword, breaking inclusion of
createrepo_c/package.h header file as C++ code:

    $ printf '#include <createrepo_c/package.h>\n' | gcc $(pkg-config --cflags createrepo_c) -c -x c++ -
    In file included from <stdin>:1:
    /usr/include/createrepo_c/package.h:118:13: error: expected unqualified-id before ‘requires’
      118 |     GSList *requires;           /*!< requires (list of cr_Dependency structs) */
          |             ^~~~~~~~

Since we support C++, there is no other way than renaming the
"requires" field of cr_Package structure.

This patch renames it to "requirements".

This patch retains original name in a schema of the SQLite database.
That should allow to reuse it on "crearerepo_c --update" after the
upgrade.

This patch retains the original name next to the new name in Python
binding, not to break Python API.

Because the rename breaks C API, major version of crearerepo_c was bumped.

Because the ABI has not changed, it decouples a shared library's
SONAME from the project version and keep it unchanged. That will save
people from rebuilding applications against the library.

(What this patch does not do is using C11 anonymous union to provide
a backward compatible name next to the new one. It would avoid
C applications from a need to adapt to the new API, yet could break
ABI because compilers are free to add a padding at the end of
a uninon.)

Resolves: #181


kontura:
I am curious what do you think about the approach with an `ifdef` (libsolv had the exact same issue https://github.com/openSUSE/libsolv/pull/477)?

kontura:
Actually I realized I would like to make the `cr_package` opaque (https://github.com/rpm-software-management/createrepo_c/issues/483)  that would also fix the issue.

ppisar:
> I am curious what do you think about the approach with an `ifdef` (libsolv had the exact same issue [openSUSE/libsolv#477](https://github.com/openSUSE/libsolv/pull/477))?

That's not a fix. If you take that libsolv header as it as and pass it to a C++20 compiler, you still get the error.

ppisar:
Replaced with #489.

PR #488, opened 2026-05-26
add support for external checksums file

M0ses:
With this patch createrepo_c will process pre-built checksum lists taken from the evironment variable `CREATEREPO_CHECKSUMS`. 
If a package was found in the list the checksum is taken from the list instead of creating the checksum of the package.
This can speed up createrepo_c significantly and reduce IO load in build environments which provide those pre-built lists.

### DIFF
--- a/src/checksum.c
+++ b/src/checksum.c
@@ -105,6 +105,32 @@ cr_checksum_name_str(cr_ChecksumType type)
     }
 }
 
+size_t
+cr_checksum_raw_size(cr_ChecksumType type)
+{
+    switch (type) {
+    case CR_CHECKSUM_UNKNOWN:
+        return 0;
+#ifdef WITH_LEGACY_HASHES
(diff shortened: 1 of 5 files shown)


dralley:
What is the functional benefit to doing this instead of using `--update`? 

There's already a pre-generated list of file checksums in the metadata itself, and it comes alongside all the rest of the metadata, which can also be re-used, rather than just the checksums.  Adding `--skip-stat` would then avoid even more I/O and processing overhead.

dralley:
Presuming you did want to continue with this approach, it would probably be better to just hash the entire pre-payload section rather than trying to hash just the main header, that way you would at least catch resigned packages.

And maybe don't use sha1 or md5 even if `LEGACY_HASHES` is enabled.  It's not much faster, and might even be slower on a modern CPU with hardware accelerated sha256

M0ses:
Thanks for your feedback

PR #489, opened 2026-05-26
2.0.0: make cr_package opaque

### LINKED ISSUE #181: Use of reserved C++ keyword

kontura:
For https://github.com/rpm-software-management/createrepo_c/issues/483
Also fixes: https://github.com/rpm-software-management/createrepo_c/issues/181

Since we would need to break API to fix the cpp `requires` keyword anyway I decided to hide everything.


dralley:
@kontura what's the timeline on 2.0 now?  I can try to help get a few things in

src/package.h
```diff
@@ -148,6 +148,608 @@ gchar *cr_package_nevra(cr_Package *package);
...
+ * @param list          changelogs list
+ */
+void cr_package_set_changelogs(cr_Package *package, GSList *list);
+
+/** Get OpenPGP signature.
+ * Ownership: not transferred.
+ * @param package       cr_Package
+ * @return              OpenPGP signature or NULL
+ */
+cr_BinaryData *cr_package_get_siggpg(cr_Package *package);
```
dralley, line 716: It doesn't have to happen in this PR, but I would maybe ditch all the functions that deal with header+payload signatures at this point.

Maybe I can add it to my PR.

kontura:
> @kontura what's the timeline on 2.0 now? 

I think that ideally it should be ready for a fedora 45 release.

> I can try to help get a few things in

That would be great.

src/python/package-py.c
```diff
@@ -412,11 +388,13 @@ set_str(_PackageObject *self, PyObject *value, void *member_offset)
...
     // inserting a new string.
-    if (!pkg->chunk)
-        pkg->chunk = g_string_chunk_new(0);
+    if (!cr_package_get_chunk(pkg))
+        cr_package_set_chunk(pkg, g_string_chunk_new(0));
 
-    char *str = PyObject_ToChunkedString(value, pkg->chunk);
-    *((char **) ((size_t) pkg + (size_t) member_offset)) = str;
+    PyObject *pybytes = PyObject_ToPyBytesOrNull(value);
+    const char *str = PyBytes_AsString(pybytes);
```
m-blaha, line 398: There should be
```
if (!pybytes) {
    return -1;
}
```
guard before calling `PyBytes_AsString(pybytes)`. The `pybytes` can be null and `PyBytes_AsString()` would in such case dereference NULL pointer: (PyBytes_AsString(NULL) => PyBytes_Check(NULL) => Py_TYPE(NULL) => NULL->ob_type).

kontura: Thanks I have added the check.

m-blaha: Hm, I think the concrete exception why unicode to utf8 conversion failed is set at this point. I'd rather not replace it and propagate the original one.

kontura: I see, ok.

m-blaha (approved):
Thank you!

PR #491, opened 2026-06-06
Delete SQLITE metadata support

dralley:
Questions:

* I removed it completely (all references to it) - should it go that far?  Does `sqliterepo_c` need to continue existing in some other package (I don't see why it would, but that would be extra conservative)
    * I also removed the error variant
* Should it go further?  e.g. sqlite metadata still could be *recognized* and then removed from repositories, given that it would no longer be kept in-sync with the XML


src/package.h
```diff
@@ -82,7 +82,7 @@ typedef struct {
 /** Package
  */
 typedef struct {
-    gint64 pkgKey;              /*!< used while inserting into sqlite db */
+    gint64 pkgKey;              /*!< package key */
```
dralley, line 85: Can this go completely?

src/xml_parser_repomd.c
```diff
@@ -51,7 +51,6 @@ typedef enum {
     STATE_SIZE,
     STATE_OPENSIZE,
     STATE_HEADERSIZE,
-    STATE_DBVERSION,
```
dralley, line 54: Should this have been removed?  Might want to keep the field and parsing code even if it doesn't get used at all.  TODO: check how createrepo_c handles tags it doesn't recognize, make sure it just skips them.

src/xml_parser_repomd.c
```diff
@@ -414,15 +411,6 @@ cr_end_handler(void *pdata, G_GNUC_UNUSED const xmlChar *element)
...
         break;
 
-    case STATE_DBVERSION:
-        assert(pd->repomd);
-        assert(pd->repomdrecord);
-
-        pd->repomdrecord->db_ver = (int) cr_xml_parser_strtoll(pd,
-                                                               pd->content,
-                                                               0);
-        break;
```
dralley, line 424: Same.  Removing this might be a little much.

src/python/locate_metadata-py.c
```diff
@@ -90,7 +87,7 @@ metadatalocation_init(_MetadataLocationObject *self,
     }
 
     /* Init */
-    self->ml = cr_locate_metadata(repopath, PyObject_IsTrue(py_ignore_db), &tmp_err);
+    self->ml = cr_locate_metadata(repopath, &tmp_err);
```
dralley, line 90: API breaking change, which is fine in a 2.0 release, just notable

src/xml_parser_repomd.c
dralley: Well, the CI failures answer that question: it breaks.

Need to add this back

kontura:
I am sorry I didn't do a proper investigation before my last comment but I found your releng ticket: https://pagure.io/releng/issue/10745 which was closed and in the last comment it states the sqlite support is still used and required.

Unfortunately I no longer think we can proceed with this.

kontura:
I have double checked and both mdapi and https://forge.fedoraproject.org/apps/fedora-packages-static do use the sqlite dbs.

dralley:
I thought that the rewrite of mdapi in Golang got rid of createrepo_c, but I guess not.

dralley:
Well, that is unfortunate.

PR #493, opened 2026-06-11
Extend the CI to test the sdist build backend

dralley:
They are slightly different - `pip install .` was working yet `python3 -m build` was not.

Resolved by removing the python module folder from .gitignore.

The `pyproject.toml` change is not relevant but it does make manual `cibuildwheel` runs much easier.


kontura (approved):
Thanks!

PR #495, opened 2026-07-15
Fix two more memory management issues

dralley:
The first commit is a minor issue in an uncommonly used API, but would be encountered in normal usage.

The second commit is defensive, but not terribly likely to be encountered in practice most of the time.


ppisar (approved):
It's good. I'll merge it.

But in the src/python/xml_parser-py.c source file there are two more calls of PyLong_FromVoidPtr() and they would also deserve a check for NULL return value. Feel free to submit a subsequent pull request.

PR #496, opened 2026-07-16
Add NULL checks for PyLong_FromVoidPtr in xml_parser callbacks

dralley:
Defensively add NULL checks to avoid issues on any potential allocation failures.

Assisted-By: Claude Opus 4.6


ppisar (approved):
Good.

PR #499, opened 2026-08-10
Correct typo s/apped/append/

### LINKED ISSUE #498: [2.0] Fix typo "cr_updateinfo_apped_record" in API
The public C function for adding advisory records to an updateinfo object is named `cr_updateinfo_apped_record` (note: "apped" instead of "append"). 

Fix this for the 2.0 release

evan-goode:
Resolves https://github.com/rpm-software-management/createrepo_c/issues/498


evan-goode:
@kontura should we keep the old spelling working and mark it as deprecated, or can we just rename it?

kontura:
We are already doing major renaming for `cr_package` without prior deprecation so I would just do it here as well.

kontura (approved):

